PR #30, opened 2018-12-16
Approve sentence if 2 out of 3 votes are valid (fixes #4)

### LINKED ISSUE #4: Define number of votes to approve a string
- Currently if a sentences has 1 positive review is approved
- Currently if a sentences has 1 negative review is rejected

We should see if we want to implement at least 2-3 votes to approve and also keep the sentences with negative votes at the end of the list just in case.

We want to avoid a single user to make the final call for sentences and avoid abuse.

MichaelKohler:


MichaelKohler:
@nukeador I went with the 2 out of 3, not sure if I understood that correctly from the meeting doc. Happy to adjust :)

nukeador:
Yes, does this allow sentences with 2 positive votes to be approved? My understanding is that if a sentence has 2 positive votes it's good to go.

MichaelKohler:
I'm merging this for now, the discussion is still ongoing. Doing fixes for the algo should be straightforward.

PR #73, opened 2019-01-08
Add missing whitespaces

davidak:
This should fix this issue.

screenshot from 2019-01-08 14-14-55

I havn't tested it.

Also, it's strange that other lines in that paragraph add whitespaces automatically. Maybe this is a general issue with the templating system. Feel free to suggest a better solution.

### DIFF
--- a/web/src/components/pages/add.jsx
+++ b/web/src/components/pages/add.jsx
@@ -312,8 +312,8 @@ const ConfirmForm = (props) => (
     )}
     <p><b>{`${props.readyCount} sentences ready for submission!`}</b></p>
     <p>
-      By submitting these sentences you grant a 
-      <a href="https://en.wikipedia.org/wiki/Public_domain" target="_blank">Public Domain License</a> 
+      By submitting these sentences you grant a {}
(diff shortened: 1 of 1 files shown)


davidak:
@MichaelKohler thanks for the feedback!

Here is a similar bugreport at https://github.com/facebook/react/issues/13696

They suggest to add ` {}`. What do you think about that? (I don't know react)

It does work, but might be not very readable.

https://codepen.io/davidak/pen/bOKzJw?editors=0010

>Move everything to one line

would make it less readable

>use a `&nbsp;` instead of the trailing whitespaces

sounds good, but 2 whitespaces get added after the link :/

https://codepen.io/davidak/pen/KbeEVy?editors=0010

MichaelKohler:
Thanks for the changes, will do a final review tonight and quickly test it locally before merging.

davidak:
good, i have updated the PR.

Use squash merge to have just one commit.

davidak:
Great. I'm waiting for the beta to actually use the tool. Keep up the good work!
I already have too many projects, so i create issues or fix them directly if trivial.

PR #92, opened 2019-01-09
Make sure we don't use a ton of GET query params for Kinto

MichaelKohler:


MichaelKohler:
For anyone reading this and thinking "WTF", yes I know this is not performant..

PR #93, opened 2019-01-09
Use separate components for SubmitForm, ReviewLink and ConfirmForm

MichaelKohler:


web/src/components/review-form.jsx
```diff
@@ -129,13 +129,13 @@ export default class ReviewForm extends React.Component {
...
-      [props.lastPage, props.lastPage + 1],
-    ].map(([ page, text ]) => (
-      <span>{
+      ['idx1', 0, '1'],
+      ['idx2', props.page - 1, '<'],
+      ['idx3', props.page, props.page + 1],
+      ['idx4', props.page + 1, '>'],
+      ['idx5', props.lastPage, props.lastPage + 1],
+    ].map(([ idx, page, text ]) => (
+      <span key={idx}>{
```
freaktechnik: Why not use the index from the array (the second map param) and make the key `key={`idx${i+1}`}`?

MichaelKohler: of course :)

PR #103, opened 2019-01-13
Feature/add auto release changelog

MichaelKohler:


MichaelKohler:
@freaktechnik r?

package.json
```diff
@@ -58,6 +64,37 @@
...
+    "semantic-release": "^15.13.3"
+  },
+  "release": {
+    "ci": false,
+    "repositoryUrl": "git@github.com:Common-Voice/sentence-collector.git",
+    "plugins": [
+      "@semantic-release/commit-analyzer",
+      "@semantic-release/release-notes-generator",
+      [
+        "@semantic-release/npm",
```
MichaelKohler: mh, I guess we could also just not use that plugin at all and that would be the same?

freaktechnik: that plugin bumps the package.json version, which is used to figure out the next version afaik.

MichaelKohler: oh, right

Issue #43, opened 2019-01-02
Add page. Explain when possible why each sentence has been rejected.


Txopi:
I added 60 sentences in Basque language and this one has been rejected: "Amerikar Estatu Batuak AEB bezala laburtzen dira." (it means something like "United States of America is shorted as USA.")

They are too many uppercase words? The problem is with the acronym? Too many words? The interface give no explanation so I don't know how to reformulate the sentence.

I understand that some regular expressions used in the source code don't have an easy explanation, but other do have (or should have). I think it deserves to make an effort and try to give a reason (at least one) why each sentence has been rejected:
- More than 10 words are not allowed
- Acronyms are not allowed
- ...

NOTE: while adding the sentences, the "Confirm" button should be disabled and the interface should say what's happening, because if they are many sentences or the server is busy, this process can take long (and users tend to click and double click all the buttons if you don't give feedback all the time...) .

MichaelKohler:
Do you think it would help for now if we linked to https://common-voice.github.io/sentence-collector/#/how-to ?

Longer term, giving exact reasons sounds reasonable.

nukeador:
Let's include a link to that page in a way it doesn't take a lot of space or distracts the workflow.

Txopi:
For the moment, it would help to link that page, but giving the exact reason next to each sentence would be really helpful for many users.

NOTE: I saw that filtered sentences are added anyway (like the one in the example above). That's what we want? That's not a list of errors, just a list of warnings?

MichaelKohler:
I've created a PR to add the link to the rejection list. 

> NOTE: I saw that filtered sentences are added anyway (like the one in the example above). That's what we want? That's not a list of errors, just a list of warnings?

Oh no, this is a huge bug. I've filed an issue to track this: https://github.com/Common-Voice/sentence-collector/issues/51

PR #127, opened 2019-01-21
Update how-to.md to add word limit

dcela:
I didn't realize there was a word limit when submitting Albanian sentences, and was wondering why they kept getting rejected.


MichaelKohler:
Thanks for filing this.

For now it's 14 words, but this will depend on the language. So I'm not sure if this makes sense to hardcode this in the HowTo.

There is also https://github.com/Common-Voice/sentence-collector/issues/43 to add a note why the sentence got rejected.

@nukeador thoughts?

nukeador:
Yes, #43 should cover that, but it's also OK to have it on the how to for now (since it's a global restriction) and then change it if we go by-language.

Issue #160, opened 2019-02-01
Improve Review Loading with more than 10k records


MichaelKohler:
Kinto has a hard-limit of 10k results. We need to check our code and decide what to do. Here are the things that just come to my mind right now:

* Export needs to export all - no limit
* The stats page on the home page only shows numbers, maybe we can improve the queries and make Kinto return the full numbers
* The review only loads 10k, which IMHO is a good thing to save bandwidth (we should probably make that even smaller), but breaks our client side sorting to show the sentences with > 0 "approval" reviews first

MichaelKohler:
The first point will be fixed in https://github.com/Common-Voice/sentence-collector/pull/171.

The second point I don't deem important right now, as 14000 would not provide a lot more value than 10000, it's mostly about showing the user that there are unreviewed sentences.

The third point I have no idea how we could fix that.

I'm renaming the issue and setting the discussion-needed flag for point 2 and 3.

dabinat:
#2 could be improved by showing rounded or less accurate numbers. For example, on the Common Voice homepage it just says 1.0k rather than listing every single hour of recording. So it could just say 10.0k+. The problem for me was more that it was implied on the homepage that 10,000 was a definitive amount.

It would also mean the stats would change less often, so there be no need to continue to generate them on every page load, which I know is another issue being discussed.

stefangrotz:
Hey I like the idea of. 10k + as a quick fix. I found the static number very confusing and with this there is at least an indicator that there are more sentences.

bact:
The review page takes too long time to load each time.

I tend to review < 100 sentences then hitting Finished Review, just to make sure that I will not lost my work if there's problem with internet connection.
But this means I have to wait for the reload again and again.

My suggestion is to limit the sentences to be loaded in each review, maybe 200 or 500 sentences max.

MichaelKohler:
This is now fixed in the latest version which just got deployed: https://discourse.mozilla.org/t/action-required-new-sentence-collector-infrastructure-and-improvements/68759

PR #171, opened 2019-02-13
chore: get all sentences (more than limit) for export (issue #160)

MichaelKohler:


MichaelKohler:
This is WIP and does not fully work, just FYI

MichaelKohler:
I've tried this with `filters.approved = true` commented out and on the production instance, but for example for Japanese, the `next` does not work as intended.

I'm getting

```
initial data totalRecords, hasNextPage 151587 true
next result { last_modified: '1548194401343',
  data: [],
  next: [Function: bound next],
  hasNextPage: false,
  totalRecords: 0 }
```

So the first call is correct, identifies that it has a next page and > 150k records. When I call `next()` I'm getting back a result, but `data` is empty and `totalRecords` is 0.

I also tried with kinto-http 4.6.1, but this doesn't change anything.

@leplatrem sorry to bother you again, do you have any idea why this is happening?

MichaelKohler:
so, there is a missing `next` update at the end of the loop. This now works against the local instance, but not against remote, but that might have another question.

MichaelKohler:
@nukeador what kinto version are we using on the prod server?

nukeador:
> @nukeador what kinto version are we using on the prod server?

Honestly I don't know (I don't have access). Nemo should be able to know.

mikehenrty:
if by production you mean https://common-voice.github.io/sentence-collector/#/,

then we are using: `9.3.0.dev0`

You can get some more information by loading this:
https://kinto.mozvoice.org/v1/

leplatrem:
> `totalRecords` is 0.

We removed the Total-Records header in the server, since it was causing troubles.
See https://github.com/Kinto/kinto-http.js/pull/323

In order to get the total number of records you must use (.getTotalRecords())[https://github.com/Kinto/kinto-http.js#total-number-of-records]

As for the `next()` function, it should work as expected, please let me know if you still encounter issues!

MichaelKohler:
Just tried this locally with 9.2.3 (as there is no 9.3.0.dev0 docker image. I still can reproduce this weirdness.

Here's what I did (on the base of this PR):

```
async getAllValidatedSentences(language) {
    const filters = {};
    // filters.approved = true;
    const collection = await this.getCollection(language);

    let { data, hasNextPage, next, totalRecords } = await collection.listRecords({ filters });
    console.log('totalRecords', totalRecords);
    while (hasNextPage) {
      const result = await next();
      console.log('totalRecords2', result.totalRecords);
      data = data.concat(result.data);
      hasNextPage = result.hasNextPage;
      next = result.next;
    }

    return data;
  }
```

Locally I get `10004` in both logs, when querying production for example for Indonesian collection I get `12885` and `0` in those logs. The same goes for the `result.data` which is empty in the second call, even though it should have data, as the limit is 10000 and there are more records than that. So somehow the next call's result does not have any data at all, but it does when running locally.

@leplatrem I don't think this has anything to do with the `Total-Records` header, or would you see a reason why that would work locally but not when querying production?

leplatrem:
> @leplatrem I don't think this has anything to do with the `Total-Records` header, or would you see a reason why that would work locally but not when querying production?

I agree. I don't see any good reason to see differences..  One idea could be to try with [the python client](https://github.com/Kinto/kinto-http.py) and see it's a client bug or not...

*untested*:
```py
import kinto_http
client = kinto_http.Client(server_url="...")
records = client.get_records(bucket="X", collection=language, pages=float('inf'), approved=True)
```

MichaelKohler:
Thanks! Now it's getting interesting, seems to be a server problem..

```
import kinto_http
# client = kinto_http.Client(server_url="http://localhost:8888/v1")
client = kinto_http.Client(server_url="https://kinto.mozvoice.org/v1")
language = "ja"
collection = "Sentences_Meta_" + language
for page in client.get_paginated_records(bucket="App", collection=collection):
  print(len(page["data"]));
```

returns `0` on the second page as well. Now, let's dive deeper into the URLs the clients actually call and see if those work in the browser..

MichaelKohler:
As far as I see `/v1/buckets/App/collections/Sentences_Meta_en/records?_limit=10000&_token=eyJsYXN0X3JlY29yZCI6eyJsYXN0X21vZGlmaWVkIjoxNTUwMDgzNTQyNzY5fSwib2Zmc2V0IjoxMDAwMCwibm9uY2UiOiJwYWdpbmF0aW9uLXRva2VuLTlkNjMxOThhLTg2MmEtNGZjMS1hZmNlLTJmZDQwM2Q1NGNmZCJ9` is being called (tested locally). The token is base64 encoded and therefore grows if there is more info encoded. However, I'd assume it's the same length for production as well.. However, there is also a possibility that we're hitting the max-character limit for the server if it's too long. I'll ask ParSys to disable the limit for now..

MichaelKohler:
This indeed has something to do with whatever we're running in front of Kinto. Using the IP directly yields all records. So this indeed is not a kinto problem. Will check with nemo/Gregor on Monday.

MichaelKohler:
I don't get it. Gregor has increased the max URL characters limit, now it still doesn't work for Japanese, Spanish or Indonesian (all of them have >20k entries). For English (over 20k as well), it responds with `10000` and `29` (which does not make any sense at all). 

All of this is using the python script from above. Still works great with the IP instead of the domain, so we circumvent nginx.

PR #176, opened 2019-02-16
Create kab.js

mozillakab:
correct punctuation style


MichaelKohler:
Thanks, I think the same also applies to English. Then it might be even better to just change the en file so kab doesn't need to be an exception?

MichaelKohler:
@mozillakab what do you think about my suggestion above?

mozillakab:
@MichaelKohler you are right. kab and en have the same rules with punctuation and capitalization. I'll put them on the en file

PR #185, opened 2019-03-02
feat: add gsw (Swiss German) language

1-Byte:
I'm not sure if this is the right place to start. I'd like to contribute Swiss German sentences. Note that there are no standardized grammar rules. Swiss German is mostly used in spoken context (instant messaging being the most notable exception) and can vary a lot depending on the region. In written form, Swiss people usually use standard German (de, de-CH).

### DIFF
--- a/shared/languages.js
+++ b/shared/languages.js
@@ -49,6 +49,11 @@ const ADDITIONAL_LANGUAGES = [
     nativeName: 'Irish',
   },
   {
+    code: 'gsw',
+    name: 'Swiss German',
+    nativeName: 'Schwizerdütsch',
+  },
+  {
     code: 'hsb',
     name: 'Upper Sorbian',
     nativeName: 'Hornjoserbšćina',


MichaelKohler:
Thanks for your efforts, I'd love to have Swiss German on it, though I'm not sure if one single language for it makes sense with all the different dialects, which all are different in writing. And given there is not a single ruleset for writing either, this might get very tricky.  I'm not sure how reliable the dataset would get. I think this is a case for linguists knowing the language to further analyze. @freaktechnik told me there are few linguists that helped the BLS to do their speech recognition at the ETH Zurich.

Let's make sure we can actually provide a valuable collection before we merge this here.

Do you want to create a topic at discourse.mozilla.org in the Common Voice category to discuss this? Then we might point a few people to it and ask for their opinions.

stefangrotz:
@1-Byte Why don't we expand this to the all allemanic speaking regions? This would include the german parts of Switzerland and the Black forest. There is the alemanic wikipedia with 26 000 articles, so one could also use the wikipedia extractor to get more sentences. https://als.wikipedia.org/wiki/Wikipedia:Houptsyte

MichaelKohler:
Thanks for creating this PR. However I'm closing this as we do not have enough information on how we would actually be able to handle all the different dialects. See also https://discourse.mozilla.org/t/deutsch-d-a-ch/46853

PR #196, opened 2019-03-18
Adding italian validator

jotaro-sama:
We added a validator for the italian language matching the [Mozilla Italia guidelines](https://github.com/Sav22999/Guide/blob/master/Mozilla%20Italia/Common%20Voice/Linee%20guida%20revisione%20Common%20Voice.md) according to which sentences for the italian dataset of Common Voice have been selected up to now.


jotaro-sama:
Creating a new one with some fixes (it worked without them on my machine but it looks like the Travis build didn't like some things).

PR #197, opened 2019-03-18
Adding italian validator (with fixes)

jotaro-sama:
We added a validator for the italian language matching the [Mozilla Italia guidelines](https://github.com/Sav22999/Guide/blob/master/Mozilla%20Italia/Common%20Voice/Linee%20guida%20revisione%20Common%20Voice.md) according to which sentences for the italian dataset of Common Voice have been selected up to now. 

Compared to the previous pull request this one has some little fixes in order to be successfully built by Travis.

### DIFF
--- a/shared/validation/languages/it.js
+++ b/shared/validation/languages/it.js
@@ -13,12 +13,12 @@ const MAX_WORDS = 17;
 // Italian: Simboli non permessi, aggiungere anche qui sotto oltre che nella regex:
 // “ ” ‘ ’ ( ) É 
 //doppio " " e più di un "." nella stessa frase.
-const SYMBOL_REGEX = /[“”‘’\(\)É(\..*\.)(  )]/       //  original:   /[<>\+\*\\#@\^\[\]\(\)\/]/;
+const SYMBOL_REGEX = /[“”‘’\(\)(\..*\.)(  )]/       //  original:   /[<>\+\*\\#@\^\[\]\(\)\/]/;
 // Any words consisting of uppercase letters or uppercase letters with a period
 // inbetween are considered abbreviations or acronyms.
 // This currently also matches fooBAR but we most probably don't want that either
 // as users wouldn't know how to pronounce the uppercase letters.
-const ABBREVIATION_REGEX = /[A-Z]{2,}|[A-Z]+\.*[A-Z]+/;
+const ABBREVIATION_REGEX = /[A-Z][a-z]{2,}|[A-Z][a-z]+\.*[A-Z][a-z]+/;
 
 
 export function getMaxLength() {
--- a/shared/validation/languages/it.js
+++ b/shared/validation/languages/it.js
@@ -13,7 +13,7 @@ const MAX_WORDS = 17;
 // Italian: Simboli non permessi, aggiungere anche qui sotto oltre che nella regex:
 // “ ” ‘ ’ ( ) É 
 //doppio " " e più di un "." nella stessa frase.
-const SYMBOL_REGEX = /[“”‘’\(\)(\..*\.)(  )]/       //  original:   /[<>\+\*\\#@\^\[\]\(\)\/]/;
+const SYMBOL_REGEX = /[“”‘’\(\)É]| {2,}|\..*\./       //  original:   /[<>\+\*\\#@\^\[\]\(\)\/]/;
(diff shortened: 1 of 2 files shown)


shared/validation/languages/it.js
```diff
@@ -0,0 +1,47 @@
...
+
+// Numbers that are not allowed in a sentence depending on the language. For
+// Italian: tutti i numeri sono consentiti ma se e quando avremo la possibilità di fare regole personalizzate metteremo una condizione ulteriore sulla lunghezza se vi sono dei numeri nella frase.
+const NUMBERS_REGEX = /[0-9]+/;
+
+/* eslint-disable-next-line no-useless-escape */
+// Italian: Simboli non permessi, aggiungere anche qui sotto oltre che nella regex:
+// “ ” ‘ ’ ( ) É 
+//doppio " " e più di un "." nella stessa frase.
+const SYMBOL_REGEX = /[“”‘’()É]| {2,}|\..*\./;       //  original:   /[<>\+\*\\#@\^\[\]\(\)\/]/;
```
MichaelKohler: can you include the missing symbols from the original as well? We certainly don't want to allow `@` or `#` for example. Whitespaces are handled in the cleanup, and if you really want to include whitespaces here, please use `/s`.

shared/validation/languages/it.js
```diff
@@ -0,0 +1,47 @@
...
+// Any words consisting of uppercase letters or uppercase letters with a period
+// inbetween are considered abbreviations or acronyms.
+// This currently also matches fooBAR but we most probably don't want that either
+// as users wouldn't know how to pronounce the uppercase letters.
+//Versione italiana: dag7dev
+const ABBREVIATION_REGEX = /[A-Z][a-z]{2,}|[A-Z][a-z]+\.*[A-Z][a-z]+/;
```
MichaelKohler: If I read this correctly (haven't tested on the website yet), this would also filter out `Test` due to the `[A-Z][a-z]{2,}` part which basically matches any word starting with a capital letter and having two following lowercase letters. Is there a reason you included `[a-z]` in this regex?

shared/validation/languages/it.js
```diff
@@ -0,0 +1,47 @@
...
+}
+
+export function getMinLength() {
+  return MIN_WORDS;
+}
+
+export function filterNumbers(sentence) {
+  //Italian: all numbers are allowed
+  var filter = !sentence.match(NUMBERS_REGEX);
+  return true || filter;
```
MichaelKohler: To ensure a good data set, we're following the requirements of the Deep Speech team, and they decided that numbers should not be allowed, independent of the language. So please use `return !sentence.match(NUMBERS_REGEX);` here.

shared/validation/languages/it.js
```diff
@@ -0,0 +1,47 @@
...
+}
+
+export function filterAbbreviations(sentence) {
+  //Italian: non abbiamo regole particolari sulle abbreviazioni
+  var filter = !sentence.match(ABBREVIATION_REGEX);
+  return true || filter;
```
MichaelKohler: To ensure a good data set, we're following the requirements of the Deep Speech team, and they decided that abbreviation should not be allowed, independent of the language. So please use `return !sentence.match(ABBREVIATION_REGEX);` here. Additionally, you have a custom abbreviation filter defined above, so this would never filter out `É` for example.

shared/validation/languages/it.js
jotaro-sama: This section was written by @dag7dev, so I leave the final decision to him, but I think his intention was to also filter abbreviations with lowercase letters. Which would be kinda hard to do in my opinion with the mechanism that was used for the english version, so it's probably better to stick to that, but let's see if our colleague has a better idea. 

In the case of abbreviations with a `.` they'll already be filtered by the symbols regex which disallows a `.` inside the sentence other than the final one.

jotaro-sama:
Just made all the requested changes (except one for which I want to wait for the other developer's feedback as described in the comment above) and also talked to the maintainer of the italian validation rules in order to change them to be fully compliant with the Deep Speech requirements.

shared/validation/languages/it.js
dag7dev: hey @MichaelKohler, as @jotaro-sama saied I wanted to filter abbreviations with lowercase. I have changed the regex as you suggested! Thanks!

PR #199, opened 2019-03-25
Update it.js

dag7dev:
Added { } according to Italian Guidelines

### DIFF
--- a/shared/validation/languages/it.js
+++ b/shared/validation/languages/it.js
@@ -10,9 +10,9 @@ const NUMBERS_REGEX = /[0-9]+/;
 
 /* eslint-disable-next-line no-useless-escape */
 // Italian: Simboli non permessi, aggiungere anche qui sotto oltre che nella regex:
-// < > + * \ # @ ^ “ ” ‘ ’ ( ) É [ ] /
+// < > + * \ # @ ^ “ ” ‘ ’ ( ) É [ ] / { }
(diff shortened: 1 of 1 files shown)


shared/validation/languages/it.js
```diff
@@ -10,9 +10,9 @@ const NUMBERS_REGEX = /[0-9]+/;
 
 /* eslint-disable-next-line no-useless-escape */
 // Italian: Simboli non permessi, aggiungere anche qui sotto oltre che nella regex:
-// < > + * \ # @ ^ “ ” ‘ ’ ( ) É [ ] /
+// < > + * \ # @ ^ “ ” ‘ ’ ( ) É [ ] / { }
 //doppio " " e più di un "." nella stessa frase.
-const SYMBOL_REGEX = /[<>+*\\#@^“”‘’()É[\]/]|\s{2,}|\..*\./;
+const SYMBOL_REGEX = /[<>+*\\#@^“”‘’()É[\]/]|\s{2,}|\..*\./{};
```
MichaelKohler: this won't work, as like this `{}` is not part of the Regex defined through the `/ ... / ` and leads to a syntax error.

dag7dev:
@MichaelKohler sorry Michael! Due to a mistake I haven't included the {} in the regex. In case of other issues tag me without problem!

Practice make the master! :D

Issue #202, opened 2019-03-28
Disallow sentences starting with lowercase letter


Adrijaned:
Would it be possible/viable to update the new sentence validator to disallow any sentences starting with lowercase letters/not ending in `[.?!]`/containing `[?!]` in the middle?

After all, sentences should be spelled correctly - thus starting with uppercase character in almost any language I can imagine. Blacklisting lowercase characters is however better choice than whitelisting uppercase due to the wide amount of scripts used around the world.

This arose mainly after having to review a great bunch of sentences that were basically probably coming from a large block of text, just randomnly split after every few words.

The ending check is because: 1) sentences should just end in one of those to be considered sentences 2) I have too reviewed a great bunch of sentences which were "Head one", "Paragraph five" and similar

The `[?!]` check in the middle is for ensuring only one sentence per line. `.` intentionally not included due to some languages using that for other things.

nukeador:
Having sentences starting with lowercase is not an issue for the deepspeech algorithm.

It seems like a too high bar to ask people to also make sure every sentence starts with upper case. Do you have other ideas to identify sentences that are no making any sense because they are not complete sentences?

MichaelKohler:
> The [?!] check in the middle is for ensuring only one sentence per line. . intentionally not included due to some languages using that for other things.

I think that definitely makes sense. I doubt there is any language that this would be valid. And if so, we can still allow it through that locale's validator if necessary. Would you like to create a PR for this? :)

jotaro-sama:
> > The [?!] check in the middle is for ensuring only one sentence per line. . intentionally not included due to some languages using that for other things.
> 
> I think that definitely makes sense. I doubt there is any language that this would be valid. And if so, we can still allow it through that locale's validator if necessary. Would you like to create a PR for this? :)


I can handle this since we already check for `.` in the middle of the sentences in the Italian validator. Do you want me to add the `[!?\.]` check in the middle only, or also to enforce the ending `[!?.]`?

Note: I will make those fully overridable in their specific language validators, since it looks like some languages could use these characters in other ways.

MichaelKohler:
> Do you want me to add the [!?\.] check in the middle only, or also to enforce the ending [!?.]?

In the middle only please.

> Note: I will make those checks fully overridable in their specific language validators, since it looks like some languages could use these characters in other ways.

Do you have examples?

jotaro-sama:
[...]

PR #203, opened 2019-03-29
Enforce multiple exclamation marks rule from the Mozilla Italia guidelines

jotaro-sama:
Simply put, I've added a check on whether there is more than one exclamation mark in a sentence, since the Italian guidelines disallow it.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "sentence-collector",
-  "version": "1.8.0",
+  "version": "1.8.2",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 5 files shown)


MichaelKohler:
Do the guidelines disallow "!!" or "test! test!"? It will only match the first one, which I think is what you want indeed. Right? The rest is probably for a discussion in https://github.com/Common-Voice/sentence-collector/issues/202

jotaro-sama:
> Do the guidelines disallow "!!" or "test! test!"? It will only match the first one, which I think is what you want indeed. Right? The rest is probably for a discussion in #202

Yes, this is what the guidelines want to disallow. The other case could probably be handled too but it's not explicitly forbidden in the rules right now, and even in that case I would handle it together with the "no multiple periods" rule.

PR #204, opened 2019-03-29
Checking whether the sentence contains [?!] in the middle

jotaro-sama:
As discussed in #202, I've added a check for those characters in the middle of a sentence as they mean it's two sentences. In the Italian version we already handled `.` in the middle so it now checks for `[?!.]`.

### DIFF
--- a/shared/validation/index.js
+++ b/shared/validation/index.js
@@ -60,6 +60,11 @@ function validateSentence(validator, sentence) {
     return validationResult;
   }
 
+  if (!validateStructure(validator, sentence)) {
+    validationResult.error = 'Contains multiple sentences';
+    return validationResult;
+  }
+
   return validationResult;
 }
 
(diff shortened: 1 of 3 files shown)


MichaelNMaggs:
Please remove those checks from the middle of sentence. It's extremely common to have [?!] in the middle, for example:

"Where are you going?" he asked.
"Help!" she shouted.

MichaelKohler:
I'll let @nukeador decide here.

nukeador:
I think we shouldn't invalidate sentences with .?! in the middle, they are correct sentences and this would bring more difficult to collect sentences.

If you still this is needed, let's open a conversation on discourse and engage with deep speech experts to understand if this is needed or not.

Thanks!

MichaelKohler:
This is reverted now (I've left it for the Italian validation though).

PR #218, opened 2019-05-13
Feature/add validator for ne locale

roshangm1:


shared/validation/languages/ne.js
```diff
@@ -0,0 +1,53 @@
...
+// < > + * \ # @ ^ [ ] ( ) /
+/* eslint-disable-next-line no-useless-escape */
+const SYMBOL_REGEX = /[<>\+\*\\#@\^\[\]\(\)\/]/;
+// Any words consisting of uppercase letters or uppercase letters with a period
+// inbetween are considered abbreviations or acronyms.
+// This currently also matches fooBAR but we most probably don't want that either
+// as users wouldn't know how to pronounce the uppercase letters.
+const ABBREVIATION_REGEX = /[A-Z]{2,}|[A-Z]+\.*[A-Z]+/;
+
+const ENGLISH_ALPHABETS_REGEX = /^[a-zA-Z]+$/;
```
MichaelKohler: this won't match anything with a space in it, or any punctuation at the end of the sentence. Here are some examples:

```
This is a sentence
This is a sentence.
This is a sentence?
ThisIsASentence.
```

roshangm1: Okay. Let me check these things.

roshangm1: I am actually not allowing english alphabets at all when nepali language is selected. Not only the sentences with space or punctuation, but sentence with english alphabets are not allowed with the current implementation. 
  I tested with other language. It's working fine. 
@MichaelKohler

MichaelKohler: Have you also tested with my example sentences and selecting Nepali? None of these match and therefore are valid. If you don't want to allow any English characters, you might use `/[a-zA-Z]/` without the `^` and `$`. This would then match always when there is an English alphabet character. Otherwise it will only match when there are no spaces or anything else, basically only if there are a-z/A-Z multiple times.

roshangm1: I got it now. Thanks for the explanation. I have made the changes accordingly. :) Tested for the above examples and few others.

roshangm1:
@MichaelKohler Anything you want to review on this?

MichaelKohler:
Will do another test tonight.

Issue #164, opened 2019-02-04
Mobile UI


dabinat:
The site is difficult to use on a phone. The text is small and a lot of manual zooming and scrolling is required.

BrunoFischerGermany:
Yes. That would be fine!

MichaelKohler:
Let's keep this simple for now:

* Let's use the existing design
* Adapt to fit to mobile where possible
* Menu could be in a hamburger menu
* All content pages are adapted to wrap correctly on mobile
* Once the sidebar is in a hamburger menu, we already should have quite a lot of screen estate saved
* The Review content is probably the most challenging.

Does somebody want to come up with an initial version for this? You can find how to set up the local environment in the README: https://github.com/Common-Voice/sentence-collector/blob/master/README.md

The CSS files can be found at ./web/css. 

Also, I'm accepting PRs for parts of this PR, it does not all need to be done in one PR. Anything better than we currently have for mobile is great! Feel free to reach out to me here if you need any help setting up / have any questions about the scope of this issue.

alejandroclose:
@MichaelKohler  I am more than happy to help with this one!!! Can you assign it to me? I'll try to have it in the next couple of weeks, ok? :)

MichaelKohler:
@alejandroclose feel free to work on it, unfortunately you'd need to be listed as contributor to this repository, which only admins can do (I can't). So I don't think it's feasible to assign bugs for now.

MichaelKohler:
See also https://github.com/Common-Voice/sentence-collector/pull/223#issuecomment-499759209

Adrijaned:
Hi, over the past few days I've thrown together a quick POC Android app for reviewing sentences, available at https://github.com/Adrijaned/CommonVoiceSentenceCollectorAndroid in case you find it useful. It is built around [Swipecards](https://github.com/Diolor/Swipecards), and I have to say at least that such a scheme of control is at least comfortable to use at mobile. I am however unsure how a such control scheme would work in the context of browser.

MichaelKohler:
@Adrijaned I haven't built the app yet, but I appreciate the efforts to explore how to make review easier. Just had a quick look at the code and noticed the following. Currently we need to do the "approval" check on the clients, so could you please implement that as well?

Reference:
https://github.com/Common-Voice/sentence-collector/blob/master/shared/db/collections/sentences-meta.js#L468
[...]

PR #223, opened 2019-06-03
Mobile UI Issue #164

alejandroclose:
@MichaelKohler Please take a look at this approach to solve Issue #164 Mobile UI: adding a toggle button to the sidebar when the screen is smaller than 480px.

Let me know what you think.
Thank you!


MichaelKohler:
@alejandroclose thanks! I will have a look tomorrow evening my time (Europe) and give my comments.

web/css/root.css
```diff
@@ -121,12 +122,74 @@ nav {
...
+}
+
+.header-btn-arrow-right {
+  border: solid white;
+  border-width: 0 3px 3px 0;
+  display: inline-block;
+  padding: 3px;
+  width: 0px;
+  margin: auto;
+  transform: rotate(-45deg);
```
MichaelKohler: minor suggestion: can we have a common base class for both arrows so we don't need to have duplicated styles between left and right? Then the definition above could be for the `header-btn-arrow` and you would only need to overwrite the transformation for the `-right` button, not setting all styles.

web/css/root.css
```diff
@@ -121,12 +122,74 @@ nav {
...
+  -webkit-transform: rotate(-45deg);
+  color: var(--second-font-color);
+}
+
+.header-btn-toggle {
+  outline: none;
```
MichaelKohler: Is there a particular reason to disable the outline? In terms of accessibility this is pretty bad as using the tab key to go from element to element won't show that the button is actually selected. Can we remove it?

web/css/root.css
```diff
@@ -121,12 +122,74 @@ nav {
...
+}
+
+/* show header button on small screens */
+
+@media screen and (min-width: 0px) and (max-width: 480px) {
```
MichaelKohler: this rule can be removed, as it has a display by default. And then we only need to hide it for bigger screens, which is done by the rule below

web/css/root.css
```diff
@@ -121,12 +122,74 @@ nav {
...
+  .header-btn-open .header-btn-closed {
+    display: block;
+  }
+}
+
+/* hide header button elsewhere */
+
+@media screen and (min-width: 601px) {
```
MichaelKohler: let's show the button for everything below 1024px, so let's change this to 1025px

web/css/root.css
```diff
@@ -121,12 +122,74 @@ nav {
...
+  }
+}
+
+nav>a {
   padding: .3rem;
   text-decoration: none;
 }
 
-nav > .active {
+nav>.active {
```
MichaelKohler: minor: can you add the spaces again here?

web/css/root.css
```diff
@@ -178,8 +241,7 @@ form input[type="checkbox"] {
   bottom: 1px;
 }
 
-form input[type="password"],
-form input[type="text"] {
+form input[type="password"], form input[type="text"] {
```
MichaelKohler: minor: can you please keep the line break here?

web/src/components/header-btn.jsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+
+export default class HeaderBtn extends React.Component {
+  render() {
+    if (this.props.isOpen) {
+      return (
+        <button className="header-btn header-btn-open" onClick={this.props.handleClick.bind(this)}>
```
MichaelKohler: given that the only thing that changes between the open/closed state is the class, can we use a conditional class here? then the class name could be generate at the beginning of the render function and we can use the same template for both cases.

web/src/components/header-btn.jsx
```diff
@@ -0,0 +1,24 @@
...
+          <div className={this.props.arrow}></div>
+          <div></div>
```
MichaelKohler: is this empty div needed?

web/src/components/header.jsx
```diff
@@ -32,4 +32,4 @@ const Header = (props) => {
   );
 };
 
-export default Header;
+export default Header;
```
MichaelKohler: can you add the empty line again at the end of the document?

alejandroclose:
Wow! Thank you @MichaelKohler !!! I will work on the changes and get back ASAP! Sorry for my rookie mistakes, hope to get better each time.

MichaelKohler:
> Sorry for my rookie mistakes, hopefully soon I'll be a better coder.

Don't worry, we all are learning :)

raivisdejus:
I am doing quite a lot of mobile reviews, two more ideas for this:

1. bigger font size on mobile for the sentences would help a lot. Say 
```.sentence-box {
    font-size: 1.5rem
}```

2. Paddings for buttons "Thumbs up", "Thumbs down" and "Finish review" could be bigger. Current size is cumbersome. Say 1.5rem?

**Comment by MichaelKohler** [comment]
@alejandroclose do you need any help here?

**Comment by alejandroclose** [comment]
@MichaelKohler I'm sorry!!! At the moment I think I'm ok, I'll work on the changes requested now to have them ASAP!

File: web/src/components/header-btn.jsx (hunk 7 above)
Comment (alejandroclose) [reply web/src/components/header-btn.jsx]: Hi @MichaelKohler , I'm a little stuck here... What do you mean by conditional class? I've been reading but I don't know if we are on the same page. Should the class name be something like `{true:'header-btn-open', false:'header-btn-closed'}`?

Comment (MichaelKohler) [reply web/src/components/header-btn.jsx]: Similar to this, yes. The syntax should be (not tested): `className={'header-btn ' + this.props.isOpen ? 'header-btn-open' : 'header-btn-closed' }`

**Comment by alejandroclose** [comment]
@MichaelKohler I believe I've made all the changes requested and fixed some other small bugs. Hope this time the PR is better! :)

File: web/src/components/header-btn.jsx [hunk 10]
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+export default class HeaderBtn extends React.Component {
+
```
MichaelKohler: minor: can you remove this line break?

web/css/root.css
```diff
@@ -176,7 +220,7 @@ form input[type="checkbox"] {
   bottom: 1px;
 }
 
-form input[type="password"],
+form input[type="password"],
```
MichaelKohler: minor: can you remove this trailing whitespace?

web/src/components/page.jsx
```diff
@@ -1,20 +1,46 @@
 import React from 'react';
 
 import Header from '../containers/header';
+import HeaderBtn from './header-btn';
+
+let DEFAULT_STATE = {
```
MichaelKohler: minor: can you use `const` here?

web/src/components/page.jsx
```diff
@@ -1,20 +1,46 @@
...
+          <main key="main">
+            {this.props.children}
+          </main>
+        </div>
+      ];
+
+    } else {
+      return [
+        <div id="page" key="page">
+          <HeaderBtn arrow="header-btn-arrow header-btn-arrow-right" handleClick={this.handleClick} />
```
MichaelKohler: as the presence of `header-btn-arrow-right` depends on `headerIsOpen`, I guess we could directly do this in `header-button.jsx`, similar to the already existing check we have there for the class name. Then we wouldn't need to pass `arrow` at all, and the `page.jsx` does not care at all if the menu is open or not in terms of styling. I think it would be good to encapsulate this, and then go further and handle the state in `header-button.jsx` as well. With these changes, `page.jsx` always has a `HeaderBtn`, but doesn't care whether it's open or not.

web/src/components/header-btn.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+export default class HeaderBtn extends React.Component {
+  
+  render() {
+      return (
+        <button className={this.props.isOpen ? "header-btn header-btn-open" : "header-btn header-btn-closed"} onClick={this.props.handleClick.bind(this)}>
```
MichaelKohler: as mentioned below, the click can directly be toggled inside this component, and then you can set the `className` of the div below according to that state.

alejandroclose:
Thank you @MichaelKohler. I'll work on the changes. I'm not sure if I can continue in this same branch and PR or should I pull everything again and continue working on a new branch & request a new PR when I'm done?

MichaelKohler:
It's fine to work on the same branch and keep pushing the requested changes to it. Then I can test and review the whole thing and merge it if all is good. Thanks for contributing! :)

MichaelKohler:
@alejandroclose just to be more clear here: you can keep working on the same branch, but you still will need to create a new PR as this one here is merged. No need to create a new branch if you don't want to though.

web/src/components/page.jsx
alejandroclose: Hi @MichaelKohler ! I managed to add the conditional for the arrow into the `header-btn.jsx`. Can you please explain a little more about the "encapsulating" part? The `Header-Btn` component should hide the header?

MichaelKohler: Depending on how you fixed the first issue, this might not be necessary anymore. Can you link me to your latest changes on GitHub so I can have another look?

MichaelKohler: (or create a PR, would work for me as well)

alejandroclose:
@MichaelKohler I did a git push to the `header-btn` branch. Let me know what you think! Thanks!!

MichaelKohler:
Currently the code looks like this:

```
render() {
    if (this.state.headerIsOpen) {
      return [
        <Header key="header" />,
        <div id="page" key="page">
          <HeaderBtn  isOpen={this.state.headerIsOpen} handleClick={this.handleClick} />
          <main key="main">
            {this.props.children}
          </main>
        </div>
      ];

    } else {
      return [
        <div id="page" key="page">
          <HeaderBtn handleClick={this.handleClick} />
          <main key="main">
            {this.props.children}
          </main>
        </div>
      ];
    }
  }
```

The only thing that changes between the if and the else is the `isOpen={this.state.headerIsOpen}`. Given that this is also exactly the condition, I would say that you can get completely rid of the if/else and just pass the condition in to `isOpen` all the times. Then you either get true or false in the component and you can handle it. So it would look like this:

```
return [
        <Header key="header" />,
        <div id="page" key="page">
          <HeaderBtn  isOpen={this.state.headerIsOpen} handleClick={this.handleClick} />
          <main key="main">
            {this.props.children}
          </main>
        </div>
      ];
```

Does that make sense?

MichaelKohler:
Also, you might not even need to return an array, this could possibly also be the following if I don't miss anything right now (not tested):

```
return (
  <Header key="header" />,
  <div id="page" key="page">
    <HeaderBtn  isOpen={this.state.headerIsOpen} handleClick={this.handleClick} />
    <main key="main">
      {this.props.children}
    </main>
  </div>
);
```

alejandroclose:
Thank you @MichaelKohler ! I believe there's another component that changes in the if/else: the  `<Header />` as it is used to show or not the header (sidebar menu). Maybe I should think of another way to make it work?

MichaelKohler:
Oh yes, of course. Let's leave that as it is for now :)

alejandroclose:
@MichaelKohler  Ok! Thank you! Please note that in this last commit to the header-btn branch I've fixed [this](https://github.com/Common-Voice/sentence-collector/pull/223#issuecomment-499759209).

MichaelKohler:
@alejandroclose do you want to create a PR for these fixes? Happy to review!

Issue #211, opened 2019-04-23
Delete all submissions from person


Adrijaned:
Through the course of reviewing past few hundred sentences, I've noticed one thing among them except for the horrible line separators that are almost random - they are all Matrix themed. However, only just now it struck to me what they really are - someone copy-pasted the whole subtitles from the Matrix movie(s?), which would explain a lot of the weird separator. That is however a huge issue, as there:

 - Almost definitely is a few hundred more of them at least, all that had to be reviewed and most of them almost definitely invalid
 - I have already approved a few of corner cases that were grammaticaly correct without realizing, which should have definitely not happened, because...
 - I am fairly certain Matrix's script is copyrighted

For reference, it's czech dataset, around 1200 pages from the end (e.g. start of sentences yet unreviewed by me), was even a lot before this, just search for `Neo`, `Seraf`, `Vědma`, `Nabukadnesar` or similar if you can

The most simple and straightforward would be to just delete all the submissions from said person, even the potentially approved ones, due to unsure licensing

TeSantana:
Hi,
It's happening the same in the Spanish language. Not with Matrix, but with a Mexican book (Angelina). It's a very old one but I've searched for information regarding the use.

I've found this information on the National Library of Spain:

_Access and/or restricted use. The present work presents restrictions regarding its reproduction, distribution, public communication or transformation; however, and pursuant to Article 37 of the Revised Text of the Intellectual Property Law (TRLPI), the National Library of Spain is authorized to reproduce and communicate works, through a closed and internal network, for the purposes of research (Article 37 , RDL 1/1996, of April 12, revised on November 5, 2015); Without prejudice to the copies that may be obtained from the BNE for private use, to exercise any other of the exploitation rights provided for in the TRLPI, the user must contact the corresponding owner or the management entity in charge of representing the owner._

I've realized about this because the sentences were appearing in an order with a lot of sense. 
If you can find the user who added these sentences probably will be necessary to delete all the submissions.

Here you have some sentences: 

"Le arrebaté las tijeras y el ovillo."
"Dame las tijeras y el ovillo."
"Mira que si no me los das no tendrás flores en tu mesa."
"Al dárselo observé que tenía los ojos arrasados en lágrimas."

Thanks

MichaelKohler:
I'll defer to @nukeador for this.

nukeador:
If these users are submitting content violating our guidelines, yes, we need to delete that.

[...]

PR #226, opened 2019-06-20
Remove mentions of Opensubtitles.org and Common Crawl from how-to

komachi:
This is very misleading links, especially with talking about how good subtitles for dataset is and then mentioning *open*subtitles, which are actually copyrighted. This sources contains mostly copyrighted content and do not provide an easy way to filter for PD content. Contributors interpret this links as a call to action and spend time to contribute sentences which we can't use. It already happened [with Swedish](https://discourse.mozilla.org/t/discussion-of-new-guidelines-for-uploaded-sentence-validation/37718/11), [Czech](https://github.com/Common-Voice/sentence-collector/issues/211#issuecomment-491522940) and now Russian.


MichaelKohler:
@nukeador do you agree with this? I've added a note yesterday, however since it's hard to filter for Public Domain parts, I can totally see this being removed.

nukeador:
Yes, let's go for it to avoid confusion.

PR #233, opened 2019-06-28
Add new fr filters

dbsaymen:


.gitignore
```diff
@@ -6,4 +6,6 @@ dist
 docker-compose.prod.yml
 .ignored_files
 .vscode
-.DS_Store
\ No newline at end of file
+.DS_Store
+.dist
+*.json
```
MichaelKohler: where is `.dist` and `.json` coming from? I doubt that we want that to exclude all json files, and there should not be a `.dist` folder..

.idea/codeStyles/Project.xml
```diff
@@ -0,0 +1,7 @@
+<component name="ProjectCodeStyleConfiguration">
```
MichaelKohler: can you please remove all the `.idea` files and add `.idea` to the .gitignore?

shared/validation/languages/fr.js
```diff
@@ -0,0 +1,53 @@
+import tokenizeWords from 'talisman/tokenizers/words';
```
MichaelKohler: how is this different from the EN file?

shared/validation/index.js
```diff
@@ -1,13 +1,14 @@
 import * as en from './languages/en';
 import * as it from './languages/it';
 import * as ne from './languages/ne';
-import * as kab from './languages/kab';
```
MichaelKohler: kab should not be removed

shared/validation/index.js
```diff
@@ -68,66 +70,67 @@ function validateSentence(validator, sentence) {
...
   }
-  
+
   if (!validateWithoutEnglishCharacters(validator, sentence)) {
     validationResult.error = 'Contains english characters';
     return validationResult;
   }
-  
+
+
```
MichaelKohler: please revert this part

shared/validation/index.js
```diff
@@ -68,66 +70,67 @@ function validateSentence(validator, sentence) {
...
 
   return validationResult;
 }
 
 function validateCorrectLength(validator, sentence) {
   const result =
-    typeof validator.filterNumbers !== 'function' ?
-      DEFAULT_VALIDATOR.filterLength(sentence) :
-      validator.filterLength(sentence);
+      typeof validator.filterNumbers !== 'function' ?
```
MichaelKohler: please revert these changes (and all below that have to additional intendation), let's keep coding style consistent

shared/validation/languages/common.js
```diff
@@ -0,0 +1,36 @@
...
+            st = st.concat(s);
+        }
+    }
+    return (st);
+}
+export function Common (sentence) {
+    // Decode any URL encoded elements of sentence
+    sentence = decode(sentence);
+    // Remove any HTML tags
+    sentence = striptags(sentence);
```
MichaelKohler: where did you see html tags? These should not even be allowed and sentences should be marked as invalid before we even come here.

shared/validation/languages/common.js
```diff
@@ -0,0 +1,36 @@
...
+    for (var s of sentence) {
+        if (allowed_categories.includes(unicode.getCategory(s.charCodeAt()))) {
+            st = st.concat(s);
+        }
+    }
+    return (st);
+}
+export function Common (sentence) {
+    // Decode any URL encoded elements of sentence
+    sentence = decode(sentence);
```
MichaelKohler: is there any reason for this?

shared/validation/languages/common.js
```diff
@@ -0,0 +1,36 @@
+var unicode = require('unicode-properties');
+var decode = require('urldecode');
+var striptags = require('striptags');
+
+function _strip_string(sentence) {
+    //letters
+    var allowed_categories = ["LC", "Ll", "Lm", "Lo", "Lt", "Lu"];
```
MichaelKohler: I don't have much experience with these categories, can you open a discussion on https://discourse.mozilla.org/c/voice so we can make sure this works for all languages?

Issue #102, opened 2019-01-13
Alert User when closing window while sentences are being uploaded


MichaelKohler:
When uploading sentences it might take some amount of time depending on the amount of sentences to be uploaded. While sentences are being uploaded, we should show a warning to the user 'beforeunload' (with `preventDefault` this is done automatically by the browser). It's best to do this through Redux, as the component might get unmounted when the user goes to a different subsite of the website. The sentences upload would still happen in the background though.

helenatxu:
Hello @MichaelKohler! I see this issue is still open, does it make sense to fix it now?

MichaelKohler:
Yes, this still makes sense. Would be great if you want to work on this once the spinner is done :)

helenatxu:
Right now I'm busy in other bugs for the projects I'm applying for the Outreachy internship, so I will unclaim that issue for now. But if nobody takes it after sometime, I will come back ;)

singhvisha:
hi want to work on this bug

singhvisha:
Hi @MichaelKohler , i want to work on this issue..

MichaelKohler:
@singhvisha sure, go ahead! I'm happy to help if you should have any questions. You can find the instructions to set up the project at https://github.com/Common-Voice/sentence-collector/blob/master/README.md

singhvisha:
hi @MichaelKohler , After installing all Prerequisites and cloning the voice-web repository locally,
when i run the frontend by running the command "yarn start",
it show this at last and nothing is happening --
E]     at processImmediate [as _immediateCallback] (timers.js:729:5)
[BE]      code: 'ECONNREFUSED',
[BE]      errno: 'ECONNREFUSED',
[BE]      syscall: 'connect',
[BE]      address: '127.0.0.1',
[BE]      port: 3306,
[BE]      fatal: true },
[BE]   isOperational: true,
[BE]   code: 'ECONNREFUSED',
[BE]   errno: 'ECONNREFUSED',
[BE]   syscall: 'connect',
[BE]   address: '127.0.0.1',
[BE]   port: 3306,
[BE]   fatal: true }
[BE] APPLICATION -- 2s to perform maintenance
[FE] Built!

MichaelKohler:
[...]

PR #247, opened 2019-08-02
feat: use history block to warn of unconfirmed sentences

HousewifeHacker:
I took a stab at https://github.com/Common-Voice/sentence-collector/issues/102 . I ran into an edge case during manual verification where if all of the sentences are filtered out, the UI and navigation flow are a little annoying. I think that's a separate issue though. This history block would start when the Submit button is clicked, and end when all the confirmation is done and state is reset. I think it would make sense to make the block start sooner, when the sentences-input has a value, but the code would be more complicated. What do you think?

### DIFF
--- a/web/src/components/pages/add.jsx
+++ b/web/src/components/pages/add.jsx
@@ -35,6 +35,22 @@ export default class Add extends React.Component {
     this.onReviewed = this.onReviewed.bind(this);
   }
 
+  componentDidMount() {
+    if (this.props.history) {
+      this.historyUnblock = this.props.history.block((nextLocation) => {
+        if (this.needsConfirmation()) {
+          return "Your sentences have not been added. Are you sure you want to leave?";
+        }
+      });
+    }
+  }
+
+  componentWillUnmount() {
+    if (this.props.history) {
+      this.historyUnblock();
+    }
+  }
+
   resetState() {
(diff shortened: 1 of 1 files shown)


MichaelKohler:
Thanks for the PR. Yes, I agree that this would be indeed if it already would block when the sentence input is filled. This would help avoid losing already written sentences. Do you want to tackle this?

MichaelKohler:
I have quickly tested this, would like to test some edge cases though. Will do that beginning of next week, thanks!

HousewifeHacker:
I was on vacation. I can spend more time on this next week or near the end of this week. I would appreciate if you can describe the edge cases so that the fix is adequate and we can add tests for them

MichaelKohler:
> I was on vacation. I can spend more time on this next week or near the end of this week. I would appreciate if you can describe the edge cases so that the fix is adequate and we can add tests for them

I wanted to test mostly around making sure it always gets unblocked when it needs to be. But thinking about it, that is indeed only after submission, before it shouldn't. I'd say all good!

Do you wanna tackle the improvement to also block when sentences are entered? Or do you wanna follow up with a second PR and we merge this one?

HousewifeHacker:
I want to tackle the improvement and test coverage.

Issue #245, opened 2019-07-30
sentences with enough vote not been exported


irvin:
It’s weird that we had add something like 4K zh-tw sentences to collector and make it review with multiple accounts, but so far we only have 1K validated and exported.

<img width="373" alt="螢幕快照 2019-07-30 下午11 37 08 1" src="https://user-images.githubusercontent.com/165936/62144180-efce1b00-b323-11e9-9265-c30527fd3962.png">

MichaelKohler:
I have looked into this the past days, for now I know the following:

1) there indeed are many records that should be approved, but the flag is not set
2) he records get updated with the username of the reviewer, so it's not a pure network error and the record just wouldn't get updated
3) I debugged this in production yesterday, and when I went through it, it counted my vote correctly and the record got marked as approved
4) it worked totally fine locally, but that was also with other data
5) therefore I assume it's an error in the check whether it should be approved, but only for some conditions, which I haven't found yet

MichaelKohler:
Irvin mentioned that they were reviewing at a meetup, and also that they used a script to pre-select the thumbs-up button. I've looked at one instance where it didn't apply the approved flag and found this:

```
"lili668668"  1554474735    Friday, April 5, 2019 2:32:15 PM
"Allen"       1553047379    Wednesday, March 20, 2019 2:02:59 AM
"irvin",      1556887501    Friday, May 3, 2019 12:45:01 PM
"tingchia",   1556887886    Friday, May 3, 2019 12:51:26 PM
"ballfish",   1556888274    Friday, May 3, 2019 12:57:54 PM
"irvin1",     1559305258    Friday, May 31, 2019 12:20:58 PM
"irvin2",     1559305893    Friday, May 31, 2019 12:31:33 PM
"ballfish1",  1561861981    Sunday, June 30, 2019 2:33:01 AM
"Lily",       1563151085    Monday, July 15, 2019 12:38:05 AM
```

If everything was in the same time, I could totally see how it would not have been approved. Currently, if somebody has an old record locally and approves, it might overwrite the approval. I heavily doubt though that all instances had old records - these would need to be at least from March 20 or older. So I don't think this is the issue at hand.

MichaelKohler:
@nukeador could you live with the fact that the export could easily take 15-20 minutes? (of course depending on network and CPU and you know it..)

[...]

PR #250, opened 2019-08-04
chore: add correct approval script

MichaelKohler:
This is a (hopefully) one-off script to correct approvals, if they have not run through the client and therefore were not correctly approved.

see #245 for context

@freaktechnik r?

### DIFF
--- a/shared/db/collections/sentences-meta.js
+++ b/shared/db/collections/sentences-meta.js
@@ -57,7 +57,7 @@ export default class SentencesMeta {
   async deleteSentenceRecords(bucket) {
     const languages = getAllLanguages();
     for (const language of languages) {
-      const records = await this.getAll(language.code);
+      const records = await this.getAllPaginated(language.code);
       const collectionName = await this.getCollectionName(language.code);
       console.log(`Found ${records.length} records to delete for ${language.code}`);
       await bucket.batch(b => {
@@ -83,7 +83,7 @@ export default class SentencesMeta {
 
   async forceDeleteSentences(bucket, locale, sentences) {
     const collectionName = await this.getCollectionName(locale);
-    const records = await this.getAll(locale);
+    const records = await this.getAllPaginated(locale);
     const minifiedRecords = records.map((record) => ({ id: record.id, sentence: record.sentence }));
     const { foundSentences, errorSentences } = sentences.reduce((acc, sentence) => {
       const foundSentence = minifiedRecords.find((record) => record.sentence === sentence);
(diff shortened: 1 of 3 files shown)


shared/db/collections/sentences-meta.js
```diff
@@ -121,6 +121,56 @@ export default class SentencesMeta {
...
+          safe: true,
+          last_modified: record.last_modified,
+        });
+      });
+    });
+
+    const adjustedResults = [];
+    const errors = [];
+    results.forEach(result => {
+      if (result.status === 200 || result === 201) {
```
freaktechnik: That seems wonky^^ Especially since the number `201` typically has no body property.

shared/db/collections/sentences-meta.js
```diff
@@ -121,6 +121,56 @@ export default class SentencesMeta {
...
+        c.updateRecord(record, {
+          safe: true,
+          last_modified: record.last_modified,
+        });
+      });
+    });
+
+    const adjustedResults = [];
+    const errors = [];
+    results.forEach(result => {
```
freaktechnik: Could use a reduce and destructure results, but this works too, I guess.

shared/db/collections/sentences-meta.js
```diff
@@ -240,6 +290,20 @@ export default class SentencesMeta {
     return result.data;
   }
 
+  async getAllPaginated(language) {
```
freaktechnik: This is essentially `getAllValidatedSentences` without the filter now, after both PRs?

MichaelKohler:
@freaktechnik re-requesting review here ;)

shared/db/collections/sentences-meta.js
```diff
@@ -332,17 +395,11 @@ export default class SentencesMeta {
...
-    const collection = await this.getCollection(language);
-
-    let { data, hasNextPage, next } = await collection.listRecords({});
-    while (hasNextPage) {
-      const result = await next();
-      data = data.concat(result.data);
-      hasNextPage = result.hasNextPage;
-      next = result.next;
-    }
+    const filters = {};
```
freaktechnik: Above the filter condition is directly in the literal.

freaktechnik:
Heh, I also see where you copied that 201 from that can't work: https://github.com/Common-Voice/sentence-collector/blob/64a48e5dfbd08db5b0c2fef7f6964b3cb95b1dbe/shared/db/collections/sentences-meta.js#L518

MichaelKohler:
Yeah, we cut the UI at 10k (first page), otherwise loading is gonna take forever. One thing why we need a backend ;)

Issue #258, opened 2019-09-22
Visual page navigation bug when submitting reviewed sentences


Adrijaned:
When displaying the "Submitting senteces, please do not close", the page nav looks like in the picture

![Screenshot_20190922_182038](https://user-images.githubusercontent.com/22203805/65391135-fb5b2280-dd65-11e9-8992-ca4dfcd03f4b.png)

MichaelKohler:
Thanks, I guess we are using a fixed width somewhere instead of being nicely responsive.

michael-dean-haynie:
@MichaelKohler, @alejandroclose - Any recent movement on this issue? I'd like to pick it up but don't want to snipe it.

MichaelKohler:
@michael-dean-haynie feel free to take this one as it's been 23 days :)

Shubhamchinda:
Hey, I'd like to work on this issue!

Adrijaned:
Hey @Shubhamchinda, I'm afraid this was already fixxed in #262 . Closing as such.

PR #262, opened 2019-10-24
fix: improve review footer layout

michael-dean-haynie:
Fix for issue #258.

~~Was able to reproduce on Firefox but not Chrome :/~~
~~On Firefox the buttons within the .pager-container have a bigger content width.~~
~~Couldn't figure out exactly why but it might be this:~~
~~https://webmasters.stackexchange.com/a/14347~~

~~Simple fix just adds a bit more buffer area to the container to account for the difference.~~

UDATE: New solution will tweak the layout and address the following issues

1. "Finish Review" button will no longer collapse
2. Pager buttons will no longer collapse no matter their length
3. "Reviews are being uploaded" message has been moved above the submit and pager buttons instead of between.


michael-dean-haynie:
Absolutely! Thanks, I'll take a look.

michael-dean-haynie:
Wow sorry that took so long.

During my testing I ended up needing to increase the min-width to 375px.
Required in Firefox while in the _"Reviews are being uploaded"_ state, with a 4-digit final pager button.

Worth noting that if the "current page" pager is big it will still break the layout:
Screenshot from 2019-12-01 12-35-21

I think to address this best would need to adjust the layout - moving that long message somewhere where it won't affect the width of the pager buttons. Not sure what the process is for suggesting layout changes though.

I also noticed that the page layout responds to changing width very differently in chrome developer tools vs the actual window. I think this is why we got different results initially.
chrome-responsive-tool
chrome-actual

MichaelKohler:
> I think to address this best would need to adjust the layout - moving that long message somewhere where it won't affect the width of the pager buttons. Not sure what the process is for suggesting layout changes though.

There is currently no official process for that. I agree with that suggestion, so I'd say let's go for it. I'd say we could move it right above the buttons and use the full width? :)

Thanks!

michael-dean-haynie:
Sounds good!

michael-dean-haynie:
Just cause it might be convenient - shots of the changes:

**While loading**
with-message

**While not loading**
no-message

PR #264, opened 2019-11-02
feat: mobile swipecards based UI

Adrijaned:
Adds a review UI mode for mobiles based on the concept of swipecards.
Basically uses https://github.com/alexandre-garrec/react-swipe-card, except its sources are directly included and somewhat edited, as I was not capable of getting it to run as is. I acknowledge that that is not an ideal solution, but consider it just a POC.
Whether the new or old UI is displayed is determined by a switch curretnly hardcoded, with no option to set from UI.
As of now, this just submits after every five sentences reviewed. That is completely fine, ecept it just fetches all the up to 10000 just after that again, and that is not good. Either add a separate submit button, or fetch only five sentences at time - though that would break the sentence sorting ATM.
This was not actually tested on mobile device due to how funky I'm connecting to my LAN. All testing was being done just in a resized desktop browser window.
Does not provide rollback functionality for now, which might eventually be deemed necessary (e.g. go to prev sentence and change how you swiped)
Does not provide any feedback as to the action taken by the swipe, some toast disaplying at least "accepted"/"rejected" would be nice to have.

### DIFF
--- a/web/css/review-form.css
+++ b/web/css/review-form.css
@@ -7,7 +7,7 @@
   padding: 0;
 }
 
-.sentence-box {
+.sentence-box, .card-sentence-box {
   height: 6rem;
   border: 1px solid var(--grey-color);
   flex-basis: 100%;
@@ -99,3 +99,37 @@
   box-shadow: none;
 }
 
+.master-root {
+  margin: 20px auto;
+  position: relative;
+  height: calc(6rem + 120px + 2em);
+  width: calc(380px + 2em);
+  overflow: hidden;
+  border: 1px solid #e5e5e5;
+}
+
+.card {
+  background-size: cover;
+  position: absolute;
+  background: #F8F3F3;
+  height: 6rem;
+  width: 250px;
+  margin: 0 auto;
+  transition: box-shadow .3s;
+  box-shadow: 0 10px 20px rgba(0,0,0,0.19), 0 6px 6px rgba(0,0,0,0.23);
+  cursor: pointer;
+  padding: 1em;
+}
+
+.animate {
+  transition: transform .3s;
+  box-shadow: none;
+}
+
+.inactive {
+  box-shadow: none;
+}
+
+.card-sentence-box {
+  border: none !important;
+}
--- a/web/css/review-form.css
+++ b/web/css/review-form.css
@@ -131,5 +131,5 @@
 }
 
 .card-sentence-box {
-  border: none !important;
+  border: none;
 }
--- a/web/css/review-form.css
+++ b/web/css/review-form.css
@@ -123,8 +123,8 @@
 .master-root {
   margin: 20px auto;
   position: relative;
-  height: calc(6rem + 120px + 2em);
-  width: calc(380px + 2em);
+  height: calc(6rem + 2em);
+  width: calc(250px + 2em);
   overflow: hidden;
   border: 1px solid #e5e5e5;
 }
@@ -135,7 +135,6 @@
   background: #F8F3F3;
   height: 6rem;
   width: 250px;
-  margin: 0 auto;
   transition: box-shadow .3s;
   box-shadow: 0 10px 20px rgba(0,0,0,0.19), 0 6px 6px rgba(0,0,0,0.23);
   cursor: pointer;
--- a/web/src/components/review-form.jsx
+++ b/web/src/components/review-form.jsx
@@ -10,7 +10,7 @@ const PAGE_SIZE = 5;
 const DEFAULT_STATE = {
   page: 0,
 };
-let mobileUI = true;
+let mobileUI = false;
 
 export default class ReviewForm extends React.Component {
   constructor(props) {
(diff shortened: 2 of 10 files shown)


MichaelKohler:
I'll have a look tomorrow :) I think this definitely is worth experimenting with. Maybe we could even do some A/B testing or something, but the 10k sentences is indeed an issue for this one. Maybe I'll have an idea.

web/src/components/review-form.jsx
```diff
@@ -38,6 +42,7 @@ export default class ReviewForm extends React.Component {
 
   async onSubmit(evt) {
     evt.preventDefault();
+    console.log("DDDDDD")
```
MichaelKohler: This could be removed ;)

web/src/components/review-form.jsx
```diff
@@ -3,11 +3,15 @@ import React from 'react';
...
 
+import Cards from './swipecard/Cards';
+import Card from "./swipecard/CardSwitcher";
+
 const PAGE_SIZE = 5;
 const DEFAULT_STATE = {
   page: 0,
   reviewed: [],
 };
+let mobileUI = true;
```
MichaelKohler: I think it would be great if this could be enabled outside of the code, no matter if it's a PoC or not. We could add that the profile as an option and then use cards if the user has chosen to do so? This could be in an "Experimental" section. What do you think?

web/src/components/review-form.jsx
```diff
@@ -98,6 +103,24 @@ export default class ReviewForm extends React.Component {
     const offset = this.getOffset();
     const curSentences = this.props.sentences.slice(offset, offset + PAGE_SIZE);
 
+    if (mobileUI) {
+      return (
+        <Cards onEnd={() => this.onSubmit({'preventDefault': function () {}})} className="master-root">
```
MichaelKohler: is the 'preventDefault' part here necessary?

web/css/review-form.css
```diff
@@ -99,3 +99,37 @@
...
+  transition: transform .3s;
+  box-shadow: none;
+}
+
+.inactive {
+  box-shadow: none;
+}
+
+.card-sentence-box {
+  border: none !important;
```
MichaelKohler: is the `important` here necessary?

web/src/components/review-form.jsx
MichaelKohler: Ah, and this would of course also need to check if we're on mobile, as we probably don't want this interface on desktop?

MichaelKohler:
I've sent the comment above too early. Here are some more comments:

1) It seems there is a merge conflict against latest master, can you fix that?
2) IMHO the sentence could fill the whole box?
3) Might be nice to have a small explanation on which direction does what :)

MichaelKohler:
I had an idea regarding the "10k reloading issue": While we can't fix this right now, as we would need a real backend for this, we could make it a bit more user friendly. We could always show a button instead of automatically submitting, however we could make that button actually useful in terms of UX. Instead of just having a "Finish Review" button that doesn't tell the user that it hasn't been submitted yet, we could have a button that says "Submit x reviews" and display that one for both cases. What do you think of that? I know, not the best case, but IMHO a worthy compromise.

web/src/components/review-form.jsx
Adrijaned: I wouldn't personally block anyone from using it on desktop. After all, things like drawing tablets exist and I even know a few people who have one. Let them use that thing for reviewing if they so wish.

MichaelKohler: Absolutely true, agreed!

web/src/components/review-form.jsx
Adrijaned: Yes, otherwise it errors, as `onSubmit` expects an object with the `preventDefault` method as its parameter. This whole thing is going to get changed to something else than onSubmit though.

web/src/components/review-form.jsx
Adrijaned: I'm afraid setting this variable from UI somewhere is beyond my webdev capabilities, largely limited to "copy and paste till it works"

Adrijaned:
1) Merge conflict fixed.
2) What do you mean by sentence filling the entire box? Could you please elaborate on that?
3) Explanation added

Just copy-pasted a button from the non-mobile interface for now, should at least work now. Default behaviour for when cards reach the end is now just loading the next page, autosubmits only when there are no more sentences to review.

web/src/components/review-form.jsx
```diff
@@ -98,6 +101,54 @@ export default class ReviewForm extends React.Component {
     const offset = this.getOffset();
     const curSentences = this.props.sentences.slice(offset, offset + PAGE_SIZE);
 
+    if (mobileUI) {
+      let message = (<p>Swipe right to approve sentencee, swipe left to reject it.</p>);
+      if (this.state.page !== 0) {
+        message = (<p>You have succesfully reviewed your {this.state.page * PAGE_SIZE}th sentence!</p>)
```
MichaelKohler: I think this should be an additional message, so the initial tutorial sentence does not get replaced

Adrijaned:
Fixed all requested; also, were it some problem to merge this not complete into `master`, feel free to push your commits to my branch first and merge it all at once, or do whatever other funky git thing.

web/src/components/review-form.jsx
```diff
@@ -98,6 +101,43 @@ export default class ReviewForm extends React.Component {
...
+                onSwipeRight={() => this.reviewSentence(offset + i, true)}
+              >
+                <div className="card-sentence-box">
+                  {sentence.sentence || sentence}
+                </div>
+              </Card>
+            ))}
+          </Cards>
+          <section className="review-footer">
+          <ConfirmButtons pendingSentences={this.state.pendingSentences}/>
```
MichaelKohler: very minor: can you indent this two more spaces?

PR #265, opened 2019-11-11
feat: add Swipe Review setting

MichaelKohler:
This adds the setting to enable the "Swiping Review" that @Adrijaned implemented in https://github.com/Common-Voice/sentence-collector/pull/264.

@freaktechnik if you want to review :)
@Adrijaned: if you want to test to make sure it works as desired

### DIFF
--- a/shared/db/collections/user.js
+++ b/shared/db/collections/user.js
@@ -81,7 +81,7 @@ export default class User {
 
     user.settings[key] = value;
     const updatedUser = await collection.updateRecord(user);
-    return updatedUser.data.languages;
+    return updatedUser.settings;
   }
 
   async addLanguage(language) {
--- a/shared/db/collections/user.js
+++ b/shared/db/collections/user.js
@@ -70,6 +70,20 @@ export default class User {
     }
   }
 
+  async setSetting(key, value) {
+    const collection = await this.getCollection();
+    const record = await collection.getRecord(this.username);
+    const user = record.data;
+
+    if (!user.settings) {
+      user.settings = {};
+    }
+
+    user.settings[key] = value;
+    const updatedUser = await collection.updateRecord(user);
+    return updatedUser.data.languages;
+  }
+
   async addLanguage(language) {
     const collection = await this.getCollection();
     const record = await collection.getRecord(this.username);
--- a/web/src/actions/index.js
+++ b/web/src/actions/index.js
@@ -21,6 +21,8 @@ export const ACTION_SUBMIT_SENTENCES_REQUEST = 'SUBMIT_SENTENCES_REQUEST';
 export const ACTION_SUBMIT_SENTENCES_SUCCESS = 'SUBMIT_SENTENCES_SUCCESS';
 export const ACTION_SUBMIT_SENTENCES_FAILURE_SINGLE = 'SUBMIT_SENTENCES_FAILURE_SINGLE';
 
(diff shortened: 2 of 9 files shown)


shared/db/collections/user.js
```diff
@@ -70,6 +70,20 @@ export default class User {
...
+    const record = await collection.getRecord(this.username);
+    const user = record.data;
+
+    if (!user.settings) {
+      user.settings = {};
+    }
+
+    user.settings[key] = value;
+    const updatedUser = await collection.updateRecord(user);
+    return updatedUser.data.languages;
```
freaktechnik: That's an interesting thing to return

shared/db/collections/user.js
```diff
@@ -70,6 +70,20 @@ export default class User {
...
+  async setSetting(key, value) {
+    const collection = await this.getCollection();
+    const record = await collection.getRecord(this.username);
+    const user = record.data;
+
+    if (!user.settings) {
+      user.settings = {};
+    }
+
+    user.settings[key] = value;
```
freaktechnik: Immutability is overrated

web/src/actions/index.js
```diff
@@ -61,6 +63,22 @@ function checkPassword(password) {
...
+  return async function(dispatch, getState) {
+    try {
+      const state = getState();
+      const db = new WebDB(state.app.username, state.app.password);
+      await db.setSetting(key, value);
+      dispatch(settingsChanged({
+        [key]: value,
+      }));
+    } catch (err) {
+      dispatch(addLanguageFailure());
```
freaktechnik: Settings are languages, suuure.

web/src/components/pages/profile.jsx
```diff
@@ -162,12 +182,26 @@ export default class Profile extends React.Component {
...
-                            filters={this.props.languages} />
-          <button disabled={this.props.pending}
+                            filters={languages} />
+          <button disabled={pending}
                   onClick={this.onAdd} className="add-language">Add</button>
         </section>
+
+        <section>
+          <h2>Settings</h2>
+          <p>Experimental: There are two different tools with which you can review sentences. The normal tool lists 5 sentences per page
```
freaktechnik: Interesting wrapping.

web/src/components/pages/profile.jsx
```diff
@@ -162,12 +182,26 @@ export default class Profile extends React.Component {
...
+          and has an approval and rejection button each. The Swiping tool displays one card at a time and you can swipe left
```
Adrijaned: Just my opinion:
```suggestion
          and has an approval and rejection button each. The Swiping tool displays one card at a time where you can swipe left
```

Adrijaned: Also, I would consider changing from swiping left __and__ right to left __or__ right (the same with approve/reject), and, changing either the order of `left and right` or `approve and reject` to match the order of actions with the order of decisions (current wording seems to imply to me you swipe left to approve and swipe right to reject)

web/src/components/pages/profile.jsx
```diff
@@ -162,12 +182,26 @@ export default class Profile extends React.Component {
...
+          and right to approve and reject sentences. Both work on Desktop, for touch interfaces we would suggest to try out
```
Adrijaned:
```suggestion
          and right to approve and reject sentences. Both are well usable on desktop, but for touch interfaces we suggest that you try out
```

PR #270, opened 2019-11-17
feat: add rm-sutsilv, rm-surmiran, rm-puter

gion-andri:
This adds the three missing Romansh (rm) idioms to the sentence collector.


MichaelKohler:
@gion-andri have you figured out in the meantime how that would fit with the new language strategy?

PR #274, opened 2019-12-12
fix(cleanup): ensure usage of unicode elipsis when it makes sense

lissyx:


MichaelKohler:
Change by itself looks good, I'm not sure though if we really should just replace all occurrences. If you, from a DeepSpeech perspective say that it should, I'd be fine with merging.

/cc @nukeador

lissyx:
> Change by itself looks good, I'm not sure though if we really should just replace all occurrences. If you, from a DeepSpeech perspective say that it should, I'd be fine with merging.
> 
> /cc @nukeador

We may want to ensure this is more consistent accross the dataset, but from a DeepSpeech perspective, it's okay imho as we will strip punctuations.

PR #276, opened 2019-12-12
chore: replace all occurences of defined issues (fixes #269)

### LINKED ISSUE #269: Cleanup does only replace first occurance
The cleanup file only replaces the first occurrences of some patterns. At least in French there are sentences with multiple of these patterns per sentence. We need to fix this, though we first need to figure out how we can avoid invalidating all Common Voice hours when we change this. That answer will be provided as an answer to https://github.com/mozilla/voice-web/issues/2259#issuecomment-548096487. 

https://github.com/Common-Voice/sentence-collector/blob/master/shared/cleanup/languages/en.js

MichaelKohler:


MichaelKohler:
I'll do a test export soon to see how many sentences this impacts. Generally it doesn't really matter as CV doesn't discard old data anyway.

test/shared/cleanup/languages/en.js
```diff
@@ -0,0 +1,35 @@
+import test from 'ava';
+import * as en from '../../../../shared/cleanup/languages/en';
+
+function clean(t, input, expected) {
+  const cleaned = en.clean(input);
+	t.deepEqual(cleaned, expected);
```
freaktechnik: indent

test/shared/cleanup/languages/en.js
```diff
@@ -0,0 +1,35 @@
...
+}
+
+test('[EN] removes multiple spaces', clean, ['this is  wrong'], ['this is wrong']);
```
freaktechnik: I don't know how [EN] is helpful when the test path already has `en` in it, but you do you.

PR #278, opened 2019-12-12
Only write json backup files if necessary

MichaelKohler:


MichaelKohler:
As discussed with Nuke we will keep these for now

PR #287, opened 2020-01-18
Create ur.js

abusarimhindi:
Adding Urdu language.
Requesting to increase number of words for sentence collector to get a meaningful sentence.
Urdu number have been added to the digits detector RegEx.

PR #288, opened 2020-01-18
Few rules of validation, have been changed for Urdu language.

abusarimhindi:


abusarimhindi:
> Good to go once #287 is approved

Updated ur.js

PR #290, opened 2020-01-19
fix(issue #123): adding sentences adds it to profile

parshva-b:
Fixed issue #123,
Assigned to me by @MichaelKohler,

Now once the user has entered his sentence, it automatically adds it to his profile

### DIFF
--- a/web/src/actions/index.js
+++ b/web/src/actions/index.js
@@ -122,8 +122,8 @@ export function submitSentences(language, sentences, source) {
     dispatch(submitSentencesSuccess(results.sentences.slice(0)));
     const errorsWithSentenceInfo = results.errors.filter((error) => error.sentence);
     dispatch(submitSentencesFailureSingle(errorsWithSentenceInfo));
-    if( !state.app.languages.includes( language ) ) {
-      dispatch(addLanguage( language ));
+    if(!state.app.languages.includes(language)) {
+      dispatch(addLanguage(language));
     }
     return results;
   };
--- a/web/src/actions/index.js
+++ b/web/src/actions/index.js
@@ -120,8 +120,12 @@ export function submitSentences(language, sentences, source) {
     const db = new WebDB(state.app.username, state.app.password);
     const results = await db.submitSentences(language, sentences, source);
     dispatch(submitSentencesSuccess(results.sentences.slice(0)));
+    dispatch( sendAddLanguage() );
     const errorsWithSentenceInfo = results.errors.filter((error) => error.sentence);
     dispatch(submitSentencesFailureSingle(errorsWithSentenceInfo));
+    if( !state.app.languages.includes( language ) ) {
+      dispatch(addLanguageSuccess(language));
+    }
     return results;
   };
 }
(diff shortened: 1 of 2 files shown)


web/src/actions/index.js
```diff
@@ -120,8 +120,12 @@ export function submitSentences(language, sentences, source) {
     const db = new WebDB(state.app.username, state.app.password);
     const results = await db.submitSentences(language, sentences, source);
     dispatch(submitSentencesSuccess(results.sentences.slice(0)));
+    dispatch( sendAddLanguage() );
     const errorsWithSentenceInfo = results.errors.filter((error) => error.sentence);
     dispatch(submitSentencesFailureSingle(errorsWithSentenceInfo));
+    if( !state.app.languages.includes( language ) ) {
+      dispatch(addLanguageSuccess(language));
+    }
```
MichaelKohler: While this adds to the profile, it won't save it in the database. You can check this by logging out and back in. Instead of having these two calls here, I'd suggest to only call `addLanguage(language)` at the end of the function which will dispatch the right messages as well as save it to the database.

web/src/reducers/index.js
```diff
@@ -112,7 +112,7 @@ function reducer(state = INITIAL_STATE, action) {
     case ACTION_ADD_LANGUAGE_SUCCESS:
       return copyInto(state, {
         pendingLanguages: false,
-        languages: action.languages,
+        languages: state.languages.concat(action.languages),
```
MichaelKohler: this shouldn't be necessary if you change it to what I suggest above.

web/src/actions/index.js
parshva-b: While testing the working of the app, I should have tested by logging out and logging back in.
I am sorry and will work to resolve this issue.
You should have a PR by day after tomorrow.

MichaelKohler: Absolutely no need to be sorry :)

parshva-b:
@MichaelKohler Have worked on the requested chances, kept the check for languages since otherwise it was trying error on saying the following:

Screenshot 2020-01-21 at 7 48 50 PM

web/src/actions/index.js
```diff
@@ -122,6 +122,9 @@ export function submitSentences(language, sentences, source) {
     dispatch(submitSentencesSuccess(results.sentences.slice(0)));
     const errorsWithSentenceInfo = results.errors.filter((error) => error.sentence);
     dispatch(submitSentencesFailureSingle(errorsWithSentenceInfo));
+    if( !state.app.languages.includes( language ) ) {
+      dispatch(addLanguage( language ));
+    }
```
MichaelKohler: Could you remove the spaces here? For this code base here, we use the following convention:

* Space after `if`, no space inside if -> `if (!state.app.languages.includes(language))`
* No spaces in parameter lists -> `addLanguage(language)`

Of course that's completely preference based :)

MichaelKohler:
Ah, also good call keeping the if statement, we of course don't want to try to add a language we already have in the profile.

web/src/actions/index.js
parshva-b: Of course, PR coming right away

PR #291, opened 2020-01-19
Add Sicilian (ISO 639-2 scn)

dpaleino:
Hello,
Sicilian (ISO 639-2 scn) has recently entered the Common Voice family. It's present in https://raw.githubusercontent.com/mozilla/voice-web/master/locales/all.json , but not in the sentence-collector webapp.

This commit should fix it -- thank you for considering it!


MichaelKohler:
Thanks, I will have a look tomorrow. I will also need to add the database table if everything is good.

PR #295, opened 2020-01-30
fix: add mapping for Hindi in exporter

phirework:
Just noticed that the CV site is fully localized in `hi-IN` but it's `hi` on sentence-collector.

### DIFF
--- a/scripts/exporter.js
+++ b/scripts/exporter.js
@@ -17,6 +17,7 @@ const LANGUAGE_MAPPING = {
   'ne-NP': 'ne',
   'sv-SE': 'sv',
   'pa-IN': 'pa',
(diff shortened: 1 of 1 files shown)


MichaelKohler:
Thanks, I will do an export tomorrow afternoon and merge this.

MichaelKohler:
Just had a quick discussion with Nuke, and I've filed https://github.com/mozilla/voice-web/issues/2550 following from that.  Let's wait for the outcome of this, weather we need to change the mapping, or we simply have "hi" in Pontoon and go from there.

MichaelKohler:
hi-IN has moved to hi, so we don't need a mapping anymore: https://github.com/mozilla/voice-web/issues/2711

PR #319, opened 2020-08-25
Add validator for Thai (th)

### LINKED ISSUE #318: Add validator for Thai (th) language
This issue is for the general discussion of what kind of filters Thai language may need for the sentence collector.

May need to keep sync with sentence extractor rules in #311  

See PR #319 for proposed code.

- filterNumbers
  - proposed: `const NUMBERS_REGEX = /[0-9๐-๙]+/;`
  - Disallowed both Arabic and Thai digits

- filterAbbreviations
  - proposed: `const ABBREVIATION_REGEX = /[A-Z]{2,}|[A-Z]+\.*[A-Z]+|[ก-ฮ]+\.([ก-ฮ]+\.)+/;`
  - Disallowed texts like AC, R.A.D., ส.ส.

- filterSymbols
  - proposed: `const SYMBOL_REGEX = /[<>+*\\#@^[\]()/\u0E4F\u0E5A\u0E5B]|[A-Za-z]{3,}/;`
  - Disallowed `// < > + * \ # @ ^ [ ] ( ) /` (from English rules)
  - Disallowed these Thai symbols:
    - Paiyannoi: \u0E2F ฯ (ellipsis, abbreviation)
    - Maiyamok: \u0E46 ๆ (repetition)
    - Fongman: \u0E4F ๏ (used as bullet)
    - Angkhankhu: \u0E5A ๚ (used to mark end of section/verse)
    - Khomut: \u0E5B ๛ (used to mark end of chapter/document)
  - Disallowed 3 or more consecutive Latin characters (can be a foreign language)

- filterStructure
  - Disallowed obvious misspelling that can be detected from character order/repetition
  - Disallowed some character orders, like tone marks cannot immediately follow the lead vowels
  - Disallowed "dangling" characters, like a lead vowel that ends a word

- filterLength
    - proposed: 2 - 110 characters
    - Use `MIN_LENGTH` and `MAX_LENGTH` (as in [Italian validator](https://github.com/Common-Voice/sentence-collector/blob/master/shared/validation/languages/it.js))
    - This supposed to keep the time length of recorded speech to be under 10 seconds

bact:
Propose code for Thai language validator.

See discussion in #318.

This will close #318.

### DIFF
--- a/shared/validation/languages/th.js
+++ b/shared/validation/languages/th.js
@@ -4,7 +4,7 @@ const MAX_LENGTH = 150;
 
 // Numbers that are not allowed in a sentence depending on the language. For
 // English this is 0-9 once or multiple times after each other.
-// Thai digits: \u0e50-\u0e59 (๐-๙)
+// Thai digits: \u0E50-\u0E59 (๐-๙)
 const NUMBERS_REGEX = /[0-9๐-๙]+/;
 
 // Some languages want to check the structure, this is what this REGEX is for. For English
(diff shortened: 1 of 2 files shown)


shared/validation/languages/th.js
```diff
@@ -0,0 +1,64 @@
+// We count chars to validate instead of words.
+const MIN_LENGTH = 2;
+const MAX_LENGTH = 150;
```
MichaelKohler, line 3: how long does it approximately take to read out a sentence with 150 characters? Of course it will be depending on the actual words etc, but looking for a guessed average here.

shared/validation/languages/th.js
```diff
@@ -0,0 +1,64 @@
...
+// - Below vowels: \u0E38\u0E39
+// - Tone marks: \u0E48\u0E49\u0E4A\u0E4B
+// - Phinthu: \u0E3A
+// - Thanthakhat: \u0E4C
+// - Nikhahit: \u0E4D
+// - Yamakkan: \u0E4E
+// The class with (*) symbol can't end the word.
+// The class with (-) symbol can't start the word.
+/* eslint-disable-next-line no-useless-escape */
+// const SYMBOL_REGEX = /[<>\+\*\\#@\^\[\]\(\)\/]|[A-Za-z]{3,}|[\u0E40\u0E41\u0E42\u0E43\u0E44]{2,}|[\u0E30]{2,}|[\u0E32\u0E33\u0E45]{2,}|[\u0E31\u0E34\u0E35\u0E36\u0E37\u0E4D\u0E47]{2,}|[\u0E38\u0E39]{2,}|[\u0E48\u0E49\u0E4A\u0E4B]{2,}|[\u0E3A]{2,}|[\u0E4C]{2,}|[\u0E4D]{2,}|[\u0E4E]{2,}|(^|\W)[\u0E30\u0E32\u0E33\u0E45\u0E31\u0E34\u0E35\u0E36\u0E37\u0E4D\u0E47\u0E38\u0E39\u0E48\u0E49\u0E4A\u0E4B\u0E3A\u0E4C\u0E4D\u0E4E]|[\u0E40\u0E41\u0E42\u0E43\u0E44]($|\W)/u;
```
MichaelKohler, line 35: what's the reason that this is still in here even though commented out? And as far as I see, the line below doesn't catch all the conditions that are mentioned above in the comment, right?

shared/validation/languages/th.js
```diff
@@ -0,0 +1,64 @@
...
+const SYMBOL_REGEX = /[<>+*\\#@^[\]()/]|[A-Za-z]{3,}|[\u0E40\u0E41\u0E42\u0E43\u0E44]{2,}|\u0E30{2,}|[\u0E32\u0E33\u0E45]{2,}|[\u0E31\u0E34\u0E35\u0E36\u0E37\u0E4D\u0E47]{2,}|[\u0E38\u0E39]{2,}|[\u0E48\u0E49\u0E4A\u0E4B]{2,}|\u0E3A{2,}|\u0E4C{2,}|\u0E4D{2,}|\u0E4E{2,}/;
```
MichaelKohler, line 36: I think many of those could very well be added to the "structure" regex, as most of that is checking structure of the sentence, not symbols. Might make it clearer.

MichaelKohler (changes requested):
Thanks for doing this! I've left a few comments. Happy to help if you have any questions.

shared/validation/languages/th.js
bact: The number is now a real guess work. Do you have a suggestion for the ideal length/range, in terms of time? 30 seconds? 20 seconds?

shared/validation/languages/th.js
bact: I have a problem using ^ and $ to catch the beginning and end of the sentence. Ideally, I would like to use \b for word boundary, but looks like regex on JavaScript doesn't very well support that for Unicode. I will remove the irrelevant comments (line 32-33) and code (35).

shared/validation/languages/th.js
bact: Good idea. Thanks for clarification of the structure regex as well. Will move.

shared/validation/languages/th.js
MichaelKohler: We're looking for around 7-10 seconds maximum.

bact: I have tried reading around 30 sentences from the sentence collector review, 7-10 seconds could be translated to 80-100 characters. I put it 110 chars to allow some more room.

shared/validation/languages/th.js
bact: Already put the character sequence checking to STRUCTURE_REGEX instead, as suggested. Thanks.

bact:
> Thanks for doing this! I've left a few comments. Happy to help if you have any questions.

@MichaelKohler Fixed all, thanks. And added some more filtering rules. 

I also put `/* eslint-disable-next-line no-misleading-character-class */` for`BEGIN_REGEX` and `END_REGEX` (part of `filterStructure`), as ESLint is picky with some Thai characters (non-based characters like Above vowels and Tone marks) when they are in the same regex with `\s', `^`, and `$`.

If that disable is discouraged, I can fallback to a rule without those characters.

MichaelKohler (approved):
Thanks!

PR #325, opened 2020-08-30
Add cleanup for Thai

### LINKED ISSUE #324: Cleanup rules for Thai
- remove
  - periods and spaces at the beginning and end of sentence
  - orphan period
  - zero-width chars (occurs in some Thai texts, due to input method or some processing)
- add
  - a space before and after Maiyamok, ?, and !
- condense
  - repetitive chars to one char: spaces, Maiyamok
- normalize
  - Sara E + Sara E -> Sara Ae
  - Nikhahit + (Tone marks +) Sara Aa -> (Tone marks +) Sara Am

--

See guidelines from the Office of the Royal Society

- `,` comma http://www.royin.go.th/?page_id=10392 (in bibliography/index use case, removing comma may change the sentence's semantic, but that use case seems irrelevant here and in any case does not affect sound)
- `:` colon http://www.royin.go.th/?page_id=10399
- `?` question mark http://www.royin.go.th/?page_id=10418
- `!` exclamation mark http://www.royin.go.th/?page_id=10433
- `ๆ` Maiyamok http://www.royin.go.th/?page_id=10427

See also validation rules #318

bact:
Cleanup code for Thai and its test set.

This will resolves #324.

### DIFF
--- a/shared/cleanup/languages/th.js
+++ b/shared/cleanup/languages/th.js
@@ -29,5 +29,6 @@ export function clean(sentences) {
       .replace(/\u0E40\u0E40/g, '\u0E41') // normalizes Sara E + Sara E -> Sara Ae
       .replace(/\u0E4d([\u0E48\u0E49\u0E4A\u0E4B]*)\u0E32/g, '$1\u0E33') // normalizes Nikhahit + Sara Aa -> Sara Am
       .replace(/([\u0E24\u0E26])\u0E32/g, '$1\u0E45') // normalizes Ru/Lu + Sara Aa -> Ru/Lu + Lakkhangyao
+      ;
   });
 }
--- a/shared/cleanup/languages/th.js
+++ b/shared/cleanup/languages/th.js
@@ -1,0 +1,26 @@
+export function sortSentences(sentences) {
+  return sentences.sort();
+}
+
+export function clean(sentences) {
+  return sentences.map((sentence) => {
+    return sentence
+      .replace(/[\u200b\u200c]/g, '')  // removes zero-width chars (occurs in some Thai texts)
+      .replace(/\?/g, ' ') // Thai does not use question mark
+      .replace(/!/g, ' ') // Thai does not use exclamation mark
+      .replace(/,/g, ' ') // Thai does not use comma
+      .replace(/(\u0E46\s*)+/g, '\u0E46') // condenses multiple Maiyamok to one Maiyamok
+      .replace(/\u0E46/g, ' \u0E46 ') // adds a space before and after Maiyamok
+      .replace(/\u0E40\u0E40/g, '\u0E41') // normalize Sara E + Sara E -> Sara Ae
+      .replace(/\u0E4d([\u0E48\u0E49\u0E4A\u0E4B]*)\u0E32/g, '$1\u0e33') // normalize Nikhahit + Sara Aa -> Sara Am
+      .replace(/\s+/g, ' ') // condenses multiple spaces to one space
+      .replace(/\s\./g, '.') // removes spaces before period
+      .replace(/\.{2,}/g, '.') // Thai does not use ellipsis (.., ...)
+      .replace(/\s:/g, ':')
+      .replace(/^\./, '') // we do not want periods at the beginning of the sentence
+      .replace(/\.$/, '') // we do not want periods at the end of the sentence
+      .replace(/^\s+/, '') // we do not want spaces at the beginning of the sentence
+      .replace(/\s+$/, '') // we do not want spaces at the end of the sentence
+      ;
+  });
+}
--- a/shared/cleanup/languages/th.js
+++ b/shared/cleanup/languages/th.js
@@ -14,20 +14,20 @@ export function clean(sentences) {
   return sentences.map((sentence) => {
     return sentence
       .replace(/[\u200b\u200c]/g, '')  // removes zero-width chars (occurs in some Thai texts)
-      .replace(/,/g, ' ') // replaces comma with space
       .replace(/:/g, ' : ') // add a space before and after colon
       .replace(/\?/g, ' ? ') // adds a space before and after question mark
       .replace(/!/g, ' ! ') // adds a space before and after exclamation mark
+      .replace(/,/g, ' ') // replaces comma with space
+      .replace(/\.(\.\s*)+/g, ' ') // replaces ellipsis (.., ...) with space
+      .replace(/\s\./g, ' ') // replaces orphan period with space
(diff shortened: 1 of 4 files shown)


MichaelKohler (approved):
Thanks, the code looks good to me. I will run a few tests and then merge it.

shared/cleanup/languages/th.js
```diff
@@ -0,0 +1,25 @@
...
+  return sentences.sort();
+}
+
+export function clean(sentences) {
+  return sentences.map((sentence) => {
+    return sentence
+      .replace(/[\u200b\u200c]/g, '')  // removes zero-width chars (occurs in some Thai texts)
+      .replace(/\?/g, ' ') // Thai does not use question mark
+      .replace(/!/g, ' ') // Thai does not use exclamation mark
+      .replace(/,/g, ' ') // Thai does not use comma
```
MichaelKohler, line 11: Replacing `!` and `?` with a space can lead to these kind of sentences, is the space there correct? Additionally this would add a space at the end of a sentence if the mark was at the end. Can you also verify that the removal of these is indeed correct? There seem to be quite a few using question marks at the end as well as inside sentences.

ไหนบอกสิเจ็บตรงไหน?
ใครขโมยพายยัดไส้ผลไม้ไป?
เธอคือนักโทษของฉันคุณรู้ไหม! 'อัศวินแดงพูดในที่สุด -> this would then include two spaces
หรือไม่สามารถทำได้, เข้าใจมั้ย -> also would then include two spaces

MichaelKohler (changes requested):

shared/cleanup/languages/th.js
bact: `.replace(/\s+/g, ' ')`
will condense multiple spaces to one

`.replace(/^\s+/, '')`
`.replace(/\s+$/, '')`
will remove spaces at the beginning and the end.

The reason I do it as a two-step process (replace those signs first with space, then condense any multiple spaces) is trying to preserve the delimiter semantic of those signs.

Consider modified example sentences here:
1) `ทำได้,เข้าใจ`
2) `ทำได้, เข้าใจ`

If you immediately remove comma (`replace(/,/g, '')`), we will get:
1a) `ทำได้เข้าใจ`
2a) `ทำได้ เข้าใจ`

If we do two-step process (`replace(/,/g, ' ')` then `replace(/\s+/g, ' ')`), we will get:
1b) `ทำได้ เข้าใจ`
2b) `ทำได้ เข้าใจ`

MichaelKohler: So, to clarify, it's okay to replace a question/exclamation mark with a space and it's still semantically and grammatically correct? It mostly seems to be an issue inside sentences, as at the end it will just be trimmed (as you correctly noted and I completely missed).

```
คุณทราบเรื่องนี้จากที่ไหน?” ฉันถาม -> คุณทราบเรื่องนี้จากที่ไหน ” ฉันถาม (I have no idea where that quote comes from, but that's how the sentence is in the DB..)
'เปลี่ยนล็อบสเตอร์อีกแล้ว!' เขาตะโกนอย่างสุดเสียง -> 'เปลี่ยนล็อบสเตอร์อีกแล้ว ' เขาตะโกนอย่างสุดเสียง
```

bact: They will still be semantically and grammatically correct.

But for semantic, it is possible in some cases that some level of semantic can be lost.

According to Office of the Royal Society, exclamation mark (!) can has three usages [1]:

1) signal an interjection (e.g. oh !)
2) signal a word that mimics natural sound (in an instance that it is impossible to written down the exact sound with available characters) (e.g. kaboom !)
3) signal an order or a warning

For question mark (?), it has two usages:

1) signal a question
2) signal a doubt (e.g. "Tailand (Thailand?)")

[1] อัศเจรีย์ (exclamation mark) http://www.royin.go.th/?page_id=10433
[2] ปรัศนี (question mark) http://www.royin.go.th/?page_id=10418 

--

For the usage (1) of question mark, which is the most common, the same sentence still retain the same semantic without the mark.

'คุณทราบเรื่องนี้จากที่ไหน?' has the same meaning as 'คุณทราบเรื่องนี้จากที่ไหน' ('ที่ไหน' at the end of this sentence means 'where', and already signalling a form of question).

There no change in sound when people read out as well, when question mark is presented in Thai, so it should be more desirable to have only one of the two sentences in the database, instead of having both.

--

For exclamation mark, removing it may affect a bit the way people read out the sentence (prosody?). So it may be reasonable to have both 'again' and 'again!' in the database.

Anyway, if that the case, I think we need to make it clear in the guideline that if people see the exclamation mark they should read it differently - like more intense, louder, etc. depends on the context.

--

Thanks for asking more about this anyway. I have also learned from the Royal Society document that, in Thai, a space is needed before ! and ?. So I will add the rules for that, if in the end we decided to keep the sign.

bact: Btw, lots of sentences have unbalanced quotes. I think we should find way to clean that automatically.

MichaelKohler: In that case, if I understand you correctly, I would suggest to not remove question and exclamation marks.

bact: Ok, then may be just normalized the spaces around the marks.

bact:
@wannaphong what do you think?

shared/cleanup/languages/th.js
wannaphong: @bact I agree with you. I think that removing question mark will not change context of the sentence in Thai language.

wannaphong:
> 
> 
> @wannaphong what do you think?

@bact I agree with you. I think that removing question mark will not change context of the sentence in Thai language.

shared/cleanup/languages/th.js
bact: I have updated the rules to be less aggressive.

They are now not removing either question mark or exclamation mark, only normalizing spaces around them (keep one space before and one space after the mark, per the Office of the Royal Society's guideline).

bact: @MichaelKohler would you mind to review this again please? We no longer remove question and exclamation marks, only add a space before and after them. Thank you.

MichaelKohler: Will have a look tomorrow!

MichaelKohler (approved):
Generally looks good to me now. Would you mind updating the test names/titles to reflect the new rules? Some of them, such as "removes space before colon" are now outdated.

bact:
Thank you. Test case names are now updated to reflect the actual thing that they test.

MichaelKohler (approved):
Thank you!

Issue #396, opened 2020-12-18
add yue (cantonese)


irvin:
Hi, reference to the discussion here (https://github.com/mozilla/common-voice/issues/2926) and the following discourse thread, please add Cantonese (yue) to sentence-collector as a new language.
https://discourse.mozilla.org/t/requesting-the-cantonese-language-yue/69928

irvin:
@MichaelKohler does it been deployed to the live site? Didn't find it at the add language dropdown list.

MichaelKohler:
@irvin retriggered the deployment, it should now show up. I thought I had verified it with the last deployment, but this makes me fairly unsure now..

irvin:
@MichaelKohler I also remember that I had check it after the issue closed. weird.

PR #397, opened 2020-12-18
add Cantonese

irvin:
add Cantonese (yue), fix https://github.com/common-voice/sentence-collector/issues/396

### DIFF
--- a/server/lib/languages.js
+++ b/server/lib/languages.js
@@ -34,7 +34,7 @@ const ADDITIONAL_LANGUAGES = [
   {
     id: 'bas',
     name: 'Basaa',
-    nativeName: 'Ɓàsàa',
+    nativeName: 'Ɓàsàa',
   },
   {
     id: 'bxr',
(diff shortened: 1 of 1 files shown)


server/lib/languages.js
```diff
@@ -34,7 +34,7 @@ const ADDITIONAL_LANGUAGES = [
   {
     id: 'bas',
     name: 'Basaa',
-    nativeName: 'Ɓàsàa',
+    nativeName: 'Ɓàsàa',
```
irvin, line 37: I had no idea why there is a diff @@

MichaelKohler:
```suggest
    nativeName: 'Ɓàsàa',
```

MichaelKohler, line 37:
```suggestion
    nativeName: 'Ɓàsàa',
```

MichaelKohler (changes requested):

MichaelKohler (approved):
Thanks!

PR #426, opened 2021-04-12
th validator: Disallow all Latin characters

bact:
Previously allow short word with 1-3 Latin characters, now disallow it all since it's clear that it's not desirable in the speech collection for Thai language.

### DIFF
--- a/server/lib/validation/languages/th.js
+++ b/server/lib/validation/languages/th.js
@@ -53,9 +53,9 @@ const END_REGEX = /[\u0E40\u0E41\u0E42\u0E43\u0E44](\s+|$)/;
 // Angkhankhu: \u0E5A ๚ (used to mark end of section/verse)
 // Khomut: \u0E5B ๛ (used to mark end of chapter/document)
 //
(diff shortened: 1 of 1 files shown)


MichaelKohler (approved):
Looks good to me, thank you!

bact:
Btw, does it possible to run the validation in retrospect with the existing sentences in the database?

We found that lots of sentences that we initially submitted are not conformed with the acceptance criteria.

MichaelKohler:
You mean sentences that are currently in the review queue in Sentence Collector? I can delete those, but can't reuse the validation file but rather need to do it through an SQL query.

bact:
I see, so could you please try to delete sentences in the queue that contain:
- Digits [0-9]
- Latin alphabets [A-Za-z]
- Thai Digits [\u0e50-\u0e59]
- Thai Mai Yamok [\u0e46]
- Thai Paiyannoi [\u0e2f]

The rest are probably too complicated. I think with just these deletes it will already helps a lot :)

Thank you again for your help.

PR #431, opened 2021-04-17
Disallow emojis for Thai sentences

bact:
- Validation will not allow emojis
- Cleanup will remove emojis

## Background

Since mid-April 2021, Mozilla Thailand Community started to add more sentences via the Sentence Collector web interface (I believe this is a preparation towards their localization event on 25 April).

One of the thing few of the members do is exporting their own Twitter archive and add them to Sentence Collector.

@wannaphong has noticed that some of the sentences have emojis. So we need to reject or remove these emojis.

### DIFF
--- a/server/lib/cleanup/languages/th.js
+++ b/server/lib/cleanup/languages/th.js
@@ -18,29 +18,29 @@ function sortSentences(sentences) {
 function clean(sentences) {
   return sentences.map((sentence) => {
     return sentence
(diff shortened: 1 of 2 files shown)


MichaelKohler:
If it helps you can run `npm run lint` locally in the server folder

bact:
> If it helps you can run `npm run lint` locally in the server folder

Thanks Michael. I should go back to my another machine that has npm environment.

MichaelKohler (approved):
Thanks

PR #441, opened 2021-05-17
[th] Fix repeating chars rule + Remove unnecessary qualifiers

bact:
- Remove unnecessary `+`s and `{n,}`s, to make the regex matches an exact number of chars (theoretically faster)
  - `/[0-9]/` will do the same job as `/[0-9]+/`, for the purpose of reject any digit in the sentence
- Fix a bug in previous no consonant repeat rule (should capture the same char repeating, not just any char in the same class). My bad.
- Add few more examples to error messages

btw, impressive new error message system - thank you.

### DIFF
--- a/server/lib/validation/languages/th.js
+++ b/server/lib/validation/languages/th.js
@@ -70,9 +70,9 @@ const INVALIDATIONS = [{
   regex: /[ก-ฮ]+\.([ก-ฮ]+\.)+/,
   error: 'Sentence should not contain abbreviations',
 }, {
-  // Five or more repeating consonants in a row is likely a non-formal spelling or difficult to read.
-  regex: /[ก-ฮ]{5,}/,
-  error: 'Sentence should not contain 5 or more consonants in a row',
+  // Seven or more repeating characters in a row is likely a non-formal spelling or difficult to read.
+  regex: /(.)\1{6,}/,
+  error: 'Sentence should not contain 7 or more of the same characters in a row',
 }, {
   regex: /[\u0E40\u0E41\u0E42\u0E43\u0E44]{2,}/,
   error: 'Sentence should not contain repeating lead vowels',
--- a/server/lib/validation/languages/th.js
+++ b/server/lib/validation/languages/th.js
@@ -41,8 +41,9 @@ const INVALIDATIONS = [{
   // Angkhankhu: \u0E5A ๚ (used to mark end of section/verse)
   // Khomut: \u0E5B ๛ (used to mark end of chapter/document)
   regex: /[<>+*\\#@^[\]()/\u0E2F\u0E46\u0E4F\u0E5A\u0E5B]/,
-  error: 'Sentence should not contain symbols',
+  error: 'Sentence should not contain symbols, including Paiyannoi and Maiyamok',
 }, {
+  // Latin character (foreign language) is not allowed
   regex: /[A-Za-z]/,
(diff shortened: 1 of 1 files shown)


server/lib/validation/languages/th.js
```diff
@@ -56,43 +56,43 @@ const INVALIDATIONS = [{
...
   // - Lead vowels
-  regex: /[\u0E40\u0E41\u0E42\u0E43\u0E44](\s+|$)/,
+  regex: /[\u0E40\u0E41\u0E42\u0E43\u0E44](\s|$)/,
   error: 'Word should not end with leading vowels',
 }, {
   // Any words consisting of letters with a period
   // inbetween are considered abbreviations or acronyms.
   // Abbreviations in Latin chars are disallowed by previous rules already.
-  regex: /[ก-ฮ]+\.([ก-ฮ]+\.)+/,
+  regex: /[ก-ฮ]\.[ก-ฮ]{1,5}\./,
```
MichaelKohler, line 70: Can you elaborate on this? Are there no abbreviations that have a longer second part than 5 characters?

server/lib/validation/languages/th.js
```diff
@@ -56,43 +56,43 @@ const INVALIDATIONS = [{
...
   error: 'Sentence should not contain abbreviations',
 }, {
-  // Five or more repeating consonants in a row is likely a non-formal spelling or difficult to read.
-  regex: /[ก-ฮ]{5,}/,
-  error: 'Sentence should not contain 5 or more consonants in a row',
+  // Seven or more repeating characters in a row is likely a non-formal spelling or difficult to read.
+  regex: /(.)\1{6}/,
```
MichaelKohler, line 74: Is that change based on new data?

server/lib/validation/languages/th.js
bact: The longest known sequence of continuous letters for any abbreviation part is 4.
- As in `รง.ซย.กรซท.ศซส.สพ.ทบ.` https://hilight.kapook.com/view/139424 

Text like `กขคงจฉช.กขคงจ.กขคงจฉ.กขคงจฉช.กขคงจฉ.` would not considered an abbreviation but more like a running text with `.` as delimiter.

This is heuristic anyway. We can have `/[ก-ฮ]\.[ก-ฮ]+\./` as well for simplicity.

MichaelKohler: As long as the rule doesn't allow sentences that are hard to read for contributors all should be fine :) As I have no idea about Thai, I'll leave that up to you.

server/lib/validation/languages/th.js
bact: - The previous regex will catch any 5+ characters in the same class.
- That is buggy, as the intention is to catch only 5+ characters of the same character.
- It should catch only: `กกกกก` and not `กฮกกฮ`.
  - So using, `( )\1` to refer to the first group that got caught.

- The number of repeating characters has changed from 5 to 7, just to be more conservative.

- `[ก-ฮ]` class of chars has also changed to `.`, to generalized the rule and reject any repeating of 7+ chars (indicator of bad quality text)

server/lib/validation/languages/th.js
bact: I committed a new code with `/[ก-ฮ]\.[ก-ฮ]+\./` for simplicity.

MichaelKohler (approved):

MichaelKohler (approved):

PR #451, opened 2021-06-12
[th] Prevent 2 separated words to stick together

bact:
Replace invalid char with a space instead of just remove it, to prevent two separated words to stick together after the removal.

Before:
- `ยิ้ม😊กิน` -> `ยิ้มกิน`

After:
- `ยิ้ม😊กิน` -> `ยิ้ม กิน`

If excessive spaces occur, they will be eliminated later.


MichaelKohler (approved):
Thanks!

PR #455, opened 2021-06-28
validation rules for Uzbek language

mkholjuraev:
We have basic validation rules here. More sophisticated rules will be introduced later after linguist experts' input

### DIFF
--- a/server/lib/validation/languages/uz.js
+++ b/server/lib/validation/languages/uz.js
@@ -5,25 +5,25 @@ const MIN_LENGTH = 1;
 const MAX_LENGTH = 125;
 
 const INVALIDATIONS = [{
-    fn: (sentence) => {
-        return sentence.length < MIN_LENGTH || sentence.length > MAX_LENGTH;
-    },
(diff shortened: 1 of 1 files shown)


server/lib/validation/languages/uz.js
```diff
@@ -0,0 +1,29 @@
...
+const MIN_LENGTH = 1;
+
+// Maximum of characters allowed per sentence to keep recordings in a manageable duration.
+const MAX_LENGTH = 125;
+
+const INVALIDATIONS = [{
+    fn: (sentence) => {
+        return sentence.length < MIN_LENGTH || sentence.length > MAX_LENGTH;
+    },
+    error: `Number of words must be between ${MIN_WORDS} and ${MAX_WORDS} (inclusive)`,
```
MichaelKohler, line 11:
```suggestion
    error: `Number of characters must be between ${MIN_LENGTH} and ${MAX_LENGTH} (inclusive)`,
```

MichaelKohler (changes requested):
Sounds like a plan, one suggestion to apply and I'll edit it to use 2 spaces instead of 4 to match with the other files.

Looking forward to future changes, thanks!

MichaelKohler (approved):

PR #457, opened 2021-07-04
Basic rule file for Esperanto

stefangrotz:
This is a basic rule file for Esperanto based on the english rule file. The only thing I changed is that I added more forbidden letters. Most important are the letters W Q X and Y that are not part of the Esperanto alphabet. But I also added many other letters that were often a problem during the Wiki Extraction.


MichaelKohler (approved):
LGTM, thanks! I will deploy this within the next few days, once done the bot will add a comment here to inform that it has been deployed.

PR #464, opened 2021-08-06
chore: fix wrong auth0 callback url in README

olejorgenb:
Had to make this change when testing locally


MichaelKohler (approved):
Indeed, thanks for fixing this!

PR #465, opened 2021-08-06
feat: request confirmation when leaving a form with unsubmitted data

olejorgenb:
I personally forgot to press the "Confirm" button on the final view a couple of
times when adding sentences. Such leave-page notifications are a bit annoying,
but more annoying to find out data has been lost IMO.


olejorgenb:
(Ref. #247 which seems to no longer be present in the code)

web/src/components/confirm-form.tsx
```diff
@@ -28,26 +29,28 @@ export default function ConfirmForm(props: Props) {
...
+      <Prompt message="Sentences not submitted, are you sure you want to leave?" when={true} />
+
       <h2>Confirm New Sentences</h2>
       <p>
         {`${submitted.length} sentences found.`}
       </p>
 
-      { invalidated.length > 0 && (
-        <p style={{color: 'red'}}>
+      {invalidated.length > 0 && (
```
MichaelKohler, line 39: Thanks for cleaning these up, those slipped in in one of my previous changes.

web/src/components/review-form.tsx
```diff
@@ -131,37 +132,42 @@ export default function SwipeReview(props: Props) {
...
-    <form id="review-form" onSubmit={onSubmit}>
+    <form id="review-form" onSubmit={ onSubmit }>
+      <Prompt
+        when={ reviewedSentencesCount > 0 }
+        message="Reviewed sentences not submitted, are sure?"
+      />
+
       <p>Swipe right to approve the sentence. Swipe left to reject it. Swipe up to skip it.</p>
-      <p>You have reviewed {reviewedSentencesCount} sentences. Do not forget to submit your review by clicking on the &quot;Finish Review&quot; button below!</p>
+      <p>You have reviewed { reviewedSentencesCount } sentences. Do not forget to submit your review by clicking on the &quot;Finish Review&quot; button below!</p>
```
MichaelKohler, line 142: Could you revert those white-space changes in this file? While these slipped in in previous changes of mine, I think we should use it without white-spaces in all cases as that's the pattern in the majority of files.

I'll definitely need to look into this (maybe Prettier nicely set up?), that's not something contributors should need to worry about.

web/src/testUtils.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
```
MichaelKohler, line 1: Could you rename this to `test-utils.tsx` and move it into the `tests` folder where we already have other files that are only concerned with test setup?

web/src/testUtils.tsx
```diff
@@ -0,0 +1,14 @@
...
+import { render } from '@testing-library/react';
+import { BrowserRouter } from 'react-router-dom';
+
+/**
+ * Render a component which expects to be inside a router
+ */
+export function renderRoute(component: React.ReactNode) {
```
MichaelKohler, line 8: While `renderWithBrowserRouter` is quite a lot longer, I'd suggest to use that naming to make it clearer on what it does. What do you think?

web/src/components/review-form.tsx
MichaelKohler: I'd be fine if you don't want to, as that is an annoying task. I can easily fix that when setting up something more automatic.

web/src/testUtils.tsx
MichaelKohler: Then maybe the comment might not even be needed anymore.

web/src/components/prompt.tsx
```diff
@@ -0,0 +1,33 @@
...
+      // Chrome requires returnValue to be set.
+      event.returnValue = '';
+    }
+
+    window.addEventListener('beforeunload', listener);
+    return () => window.removeEventListener('beforeunload', listener);
+  }, [when]);
+
+  return (
+    <ReactRouterPrompt {...props}/>
```
MichaelKohler, line 31: As far as I can see we could directly use `Prompt` from `react-router-dom` in the components, as all the event listener set up is already handled by it. Was there a specific reason why you had to wrap it with this component?

MichaelKohler (changes requested):
Thanks for this contribution! It's great to see somebody contribute to the Sentence Collector after quite a while! While I don't particularly like leave-page prompts either, I think these make sense in this case.

> (Ref. #247 which seems to no longer be present in the code)

I did quite a big infrastructure refactoring. Previously sentences were uploaded quite slowly and that prompt helped a lot to make sure that users don't close the tab while it's still uploaded. With the new infrastructure uploads happen way faster and I removed it. However I did not think about the use case where the upload has not even happened yet, that definitely was a mistake, thanks for catching that.

I've left a few comments, happy to provide more context where needed!

Did you stumble upon something else that could make contributions easier? I've already merged your other PR regarding the wrong callback in the README. Anything else that would have made your contribution easier?

Thanks again!

web/src/components/review-form.tsx
MichaelKohler: Don't worry about those, I've created https://github.com/common-voice/sentence-collector/pull/466 which will take care of that once this is merged :)

web/src/components/confirm-form.tsx
olejorgenb: Including a editorconfig file (IIRC common-voice does, but not sentence-collector) would help slightly, but unfortunately the options seems limited to basic indentation, not language specific styles such as this).

These crept in after indenting and unindenting some lines. I think I correct most of them.

web/src/components/prompt.tsx
olejorgenb: I'm not sure I follow - the `Prompt` from `react-router-dom` does not handle `beforeunload` (https://github.com/ReactTraining/react-router/pull/4372 was never merged as far as I can see). I did tested and no warning was issued when simply closing the tab.

web/src/testUtils.tsx
olejorgenb: > Did you stumble upon something else that could make contributions easier? I've already merged your other PR regarding the wrong callback in the README. Anything else that would have made your contribution easier?

Not really (apart from my default editor formatting not matching the project style). While there was a couple extra steps compared to common-voice, the setup was really painless and worked out of the box :)

web/src/components/prompt.tsx
MichaelKohler: Right, I should not review PRs before my second coffee kicks in. Thanks for the elaboration!

MichaelKohler (approved):
LGTM! Thanks a lot for this contribution.

PR #472, opened 2021-09-04
add nan-TW (Taiwanese) language

irvin:
https://github.com/common-voice/common-voice/issues/3194


MichaelKohler (approved):
Thanks, I will double check this on staging and then push it to production

PR #482, opened 2021-09-15
Referencing + CC0 waiver process

Heyhillary:
update to how to include how to reference and cc0 waiver process


MichaelKohler (approved):
Thanks! I will update the tests in a followup.

Issue #487, opened 2021-10-10
Enabling uppercase for key based validation


HarikalarKutusu:
I think it is in web/src/components/review-form.tsx... It only handles lowercase letters. I hit this while reviewing with caps-lock on. Maybe it is a good idea to lowercase it before checks...

MichaelKohler:
Indeed, the code block in https://github.com/common-voice/sentence-collector/blob/main/web/src/components/review-form.tsx#L111-L123 should first convert it to lowercase and then do these checks. Nice find!

PR #508, opened 2021-10-31
Fix for uppercase keys not working #487

HarikalarKutusu:
This is a fix for https://github.com/common-voice/sentence-collector/issues/487
This had been removed from localization PR.


MichaelKohler (approved):
Thanks!

PR #527, opened 2021-11-06
chore: add localization documentation (fixes #524)

### LINKED ISSUE #524: Document localization process for Sentence Collector
As per decision in #491 we need to document the flow and how everything fits together. Otherwise it will be confusing for contributors to Sentence Collector as the localization setup is not standard.

MichaelKohler:


MichaelKohler:
@HarikalarKutusu  does https://github.com/common-voice/sentence-collector/blob/feature/localization-docs/docs/LOCALIZATION.md make sense to you?

HarikalarKutusu:
> @HarikalarKutusu does https://github.com/common-voice/sentence-collector/blob/feature/localization-docs/docs/LOCALIZATION.md make sense to you?

It is perfect!

One (possibly unrelated) question thou: As this is an open-source project, one can install it locally or on their own servers, not for development, but for actual use. The planned workflow will prevent localized content in such instances.

I'm just being picky :)

On the other hand, putting the strings in the CV project in Pontoon is a good call because there already exist working groups for translation.

MichaelKohler:
Thanks for the feedback.

> One (possibly unrelated) question thou: As this is an open-source project, one can install it locally or on their own servers, not for development, but for actual use. The planned workflow will prevent localized content in such instances.

Not necessarily, you could still just add the translated files in the `web/locales` folder in your own fork and use your own localization setup with that structure. Or even do the same step as this repository will do downloading the same translations when deploying.

HarikalarKutusu:
Minor copy-paste issue:
"Deleting a new string" => "Deleting a string"

PR #529, opened 2021-11-07
adding js rules for Igbo

chrisemezue:
@MichaelKohler  sorry i am doing it late. Here it is for Igbo.


MichaelKohler (approved):
Thanks! I've added a small change to add the validator to the validator list as well.

PR #544, opened 2021-11-17
Localization: Fix for Pontoon related issue #543

### LINKED ISSUE #543: Issue in language strings in Pontoon
For readability purposes, we divided the sentences into lines. Such as:

```
sc-home-intro = %The Sentence Collector is part of <commonVoiceLink>Common Voice</commonVoiceLink>.
    It allows contributors to collect and validate sentences created by the community.
    You can use this tool also to import and clean-up small-to-medium-sized public domain corpus 
    you have found or collected. All sentences need to be Public Domain.
    Approved sentences are exported every week to the Common Voice repository and are released
    on the Common Voice website on every new deployment.
```

But when it is transferred to Pontoon, they are also regarded as different lines.
So, for example, Google Translate helper gets confused from that. It tries to translate the following separately:

1. You can use this tool also to import and clean up small-to-medium-sized public domain corpus
2. you have found or collected. All sentences need to be Public Domain.

Of course, we shouldn't rely on Google Translate, but it gives us a head start... We copy it and correct it...

I'm preparing a PR now...  Another export would be advisable before the sentences got translated.

HarikalarKutusu:
This fixes #543 by simply eliminating extra lines in the ftl file.

### DIFF
--- a/web/locales/en/messages.ftl
+++ b/web/locales/en/messages.ftl
@@ -27,12 +27,7 @@ sc-logout-button = Logout
 ###
 
 sc-home-title = Welcome to the Common Voice Sentence Collector
-sc-home-intro = The Sentence Collector is part of <commonVoiceLink>Common Voice</commonVoiceLink>.
-    It allows contributors to collect and validate sentences created by the community.
-    You can use this tool also to import and clean-up small-to-medium-sized public domain corpus
-    you have found or collected. All sentences need to be Public Domain.
-    Approved sentences are exported every week to the Common Voice repository and are released
-    on the Common Voice website on every new deployment.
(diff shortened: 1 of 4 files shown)


MichaelKohler:
Thanks for the PR and nice catch on the issue! This looks good to me apart from the failing tests. I quickly looked through the diffs in the snapshots and they look good to me. Do you have time to update the snapshots to reflect the latest changes so subsequent tests will have the correct baseline to compare to? You could do that by running `npm t -- -u` and committing the changes. If you don't have time for it, then I could also take over, no biggie.

MichaelKohler (approved):
Thanks!

PR #546, opened 2021-11-18
Localization: Several typo/wording corrections in ftl for review

HarikalarKutusu:
Changes can be seen in the commit. One is more important than others, placement of the button...
Nothing breaking...


web/locales/en/messages.ftl
```diff
@@ -218,7 +218,7 @@ sc-criteria-item-1 = The sentence must be spelled correctly.
 sc-criteria-item-2 = The sentence must be grammatically correct.
 sc-criteria-item-3 = The sentence must be speakable.
 sc-criteria-item-4 = If the sentence meets the criteria, click the &quot;Approve&quot; button on the right.
-sc-criteria-item-5 = If the sentence does not meet the above criteria, click the &quot;Reject&quot; button on the right. If you are unsure about the sentence, you may also skip it and move on to the next one.
+sc-criteria-item-5 = If the sentence does not meet the above criteria, click the &quot;Reject&quot; button on the left. If you are unsure about the sentence, you may also skip it and move on to the next one.
```
MichaelKohler, line 221: I feel this would warrant triggering a re-translation. What do you think? If you agree, let's change the key (both here and where it's used in the component) to trigger that. Firefox is using `-2` prefixes if there are no clear other changes, so I'd go with `sc-criteria-item-5-2`, as odd as it is.

web/locales/en/messages.ftl
```diff
@@ -47,8 +47,8 @@ sc-howto-addsen-post-1 = We prefer natural/conversational sentences. While phone
 sc-howto-cite-title = How to Cite
 sc-howto-cite-pre-1 = It's important that you reference where you found the public licence to avoid plagiarism and to allow for follow ups on <copyrightIssuesLink>copyright issues</copyrightIssuesLink>. Here are a few ways you can cite the work.
 sc-howto-cite-item-1 = You could include the website, e.g "Common Voice - https://commonvoice.mozilla.org/"
-sc-howto-cite-item-2 = You could use Academic referencing style, e.g harvard style "Mozilla (2021) Common Voice. Available at https://commonvoice.mozilla.org/ (Accessed: 15th September 2021)"
-sc-howto-cite-item-3 = For public licence text not available online, you could use Academic referencing style e.g harvard style "Jess (2021) My Public licence poems"
+sc-howto-cite-item-2 = You could use academic referencing style, e.g Harvard style "Mozilla (2021) Common Voice. Available at https://commonvoice.mozilla.org/ (Accessed: 15th September 2021)"
+sc-howto-cite-item-3 = For public licenced text not available online, you could use academic referencing style e.g Harvard style "Jess (2021) My Public licence poems"
```
MichaelKohler, line 51: Without being a lawyer, editor or copywriter, I think "public license text" would be better than your suggestion. I feel "licenced" introduces some ambiguity. Maybe "For public domain texts not available online .." would be more descriptive?

MichaelKohler (changes requested):
Thanks! Two comments, the rest looks good to me.

web/locales/en/messages.ftl
HarikalarKutusu: Should be OK now.

web/locales/en/messages.ftl
HarikalarKutusu: Changed to "public domain text"...

MichaelKohler (approved):
LGTM, thanks!

Issue #567, opened 2021-11-25
Link on logo in header reverts back UI to English


HarikalarKutusu:
To reproduce:

1. Change UI language in Profile to another language => Works OK for other links...
2. Click logo to get to homepage => Falls back to English

This part in https://github.com/HarikalarKutusu/sentence-collector/blob/main/web/src/components/header.tsx (Line 69)
```js
        <Link to="/" href="">
          <img src={logoURL} />
        </Link>
```
should also use useLocaleUrl() ...

MichaelKohler:
Thanks for filing this. For the Home link in the navigation we set the correct link: https://github.com/common-voice/sentence-collector/blob/4abc875c06ed111a48a5d4b4f5fbbbb3b690fb4d/web/src/components/header.tsx#L25

But we don't do the same with the logo link at the top left of the header, yes. We should use that exact same approach for it as well using `useLocaleUrl`: https://github.com/common-voice/sentence-collector/blob/4abc875c06ed111a48a5d4b4f5fbbbb3b690fb4d/web/src/components/header.tsx#L69

venkateshprabhu2:
@MichaelKohler Can I work on this? How do I get started?

MichaelKohler:
Yes, this is free to be worked on. Here's the approach I would suggest:

* Set up the local environment as outlined in the [README](https://github.com/common-voice/sentence-collector/blob/main/README.md#local-development)
* Look at the issue description, make sure you can reproduce the issue locally and fully understand the issue
* Implement what is needed
* Test it locally and verify that your changes fix the issue
* Make sure that the automated tests pass locally (see the README for how to run the tests)

If you have any questions and run into issues, feel free to either post here or join our Matrix chat.

venkateshprabhu2:
I setup my local env and am trying to reproduce the bug. When I click on login, I am getting an oauth error. I have posted the same in matrix chat as well. Can you help on how I can resolve it further? 

`request to https://pontoon.mozilla.org/graphql failed, reason: unable to get local issuer certificate',InternalOAuthError: Failed to obtain access token`

PR #581, opened 2022-01-13
fix: link on logo in header reverts back UI to English

venkateshprabhu2:
Fix for https://github.com/common-voice/sentence-collector/issues/567


MichaelKohler:
Thank you! Works perfectly well.

PR #605, opened 2022-02-13
Add sentence validator for Cantonese

laubonghaudoi:


server/lib/validation/languages/yue.js
```diff
@@ -0,0 +1,34 @@
+// Minimum of words that qualify as a sentence.
```
MichaelKohler, line 1: minor: could you please change the comment here as it's now number of characters and not words?

server/lib/validation/languages/yue.js
```diff
@@ -0,0 +1,34 @@
...
+const MIN_LENGTH = 3;
+
+// Maximum of words allowed per sentence to keep recordings in a manageable duration.
```
MichaelKohler, line 4: minor: could you please change the comment here as it's now number of characters and not words?

server/lib/validation/languages/yue.js
```diff
@@ -0,0 +1,34 @@
...
+  },
+  error: `Number of characters must be between ${MIN_LENGTH} and ${MAX_LENGTH} (inclusive)`,
+}, {
+  regex: /[0-9]+/,
+  error: "Sentence should not contain numbers",
+}, {
+  regex: /[<>+*#@%^[\]()\/]/,
+  error: "Sentence should not contain symbols",
+}, {
+  // 7 or more repeating characters in a row is likely a non-formal spelling or difficult to read.
```
MichaelKohler, line 19: Interesting, did that happen often in Sentence Collector? 7 repeating characters seems like a lot, but I have absolutely no language experience apart from latin-based languages.

server/lib/validation/languages/yue.js
```diff
@@ -0,0 +1,34 @@
...
+  regex: /(.)\1{6}/,
+  error: "Sentence should not contain more than 7 of the same character in a row",
+}, {
+  // Emoji range from https://www.regextester.com/106421 and
+  // https://stackoverflow.com/questions/10992921/how-to-remove-emoji-code-using-javascript
+  regex: /(\u00a9|\u00ae|[\u2000-\u3300]|[\u2580-\u27bf]|\ud83c[\ud000-\udfff]|\ud83d[\ud000-\udfff]|\ud83e[\ud000-\udfff]|[\ue000-\uf8ff])/,
+  error: "Sentence should not contain emojis or other special Unicode symbols",
+}, {
+  regex: /[\u5427](\s|$)/,
+  error: 'Sentence should not end with Mandarin particles',
```
MichaelKohler, line 29: Is this message correct? This would also reject `\u5427` followed by a space. Is that also considered ending a sentence?

MichaelKohler (changes requested):
Thanks for working on this. I've added a few minor comments.

Additionally to those: I think it would be beneficial to translate the error messages. That was for example done here: https://github.com/common-voice/sentence-collector/blob/main/server/lib/validation/languages/ckb.js . We decided not to translate these error messages through the normal UI translation process, as that would require quite some work for not much benefit. However if somebody is adding sentences in a given language, it's likely they would understand a translated error message anyway. Translating those error messages would then enable contributors to contribute without knowing English. Do you think that would be a good improvement here?

server/lib/validation/languages/yue.js
laubonghaudoi: This sometimes happens when people dump uncleaned sentences directly crawled from web. For examples sentences with long tailing dots, such as`額........` Such sentences are most likely junk.

server/lib/validation/languages/yue.js
laubonghaudoi: Yes sometimes a space indicates a pause or the end of a sentence. I have amended this rule in the latest commit.

laubonghaudoi:
Thanks for reviewing! It is now fixed in the latest commit.

server/lib/validation/languages/yue.js
```diff
@@ -0,0 +1,34 @@
...
+const INVALIDATIONS = [{
+  fn: (sentence) => {
+    return sentence.length < MIN_LENGTH || sentence.length > MAX_LENGTH;
+  },
+  error: `字數必須要喺 ${MIN_LENGTH} 同  ${MAX_LENGTH} 之間`,
+}, {
+  regex: /[0-9]+/,
+  error: "句子唔可以包含阿拉伯數字",
+}, {
+  regex: /[<>+*#@%^[\]()\/]/,
```
MichaelKohler, line 16:
```suggestion
  regex: /[<>+*#@%^[\]()/]/,
```

MichaelKohler (commented):
Thanks for working on this. One more small comment that I didn't catch before and then this can get merged :)

MichaelKohler (dismissed):
Thanks!

MichaelKohler (approved):

MichaelKohler:
This will be part of the next release. I can't say yet when that will be, but there will be a new comment once the release is done.

laubonghaudoi:
Great, thank you so much!

PR #606, opened 2022-02-19
feat: add sentence validator for catalan

jmigual:
Uses the gersam tokenizer for Catalan that counts a word with a dash as a single word.

### DIFF
--- a/server/lib/validation/languages/ca.js
+++ b/server/lib/validation/languages/ca.js
@@ -1,12 +1,17 @@
-// According to Mozilla Italia guidelines, we count chars to validate instead of words.
-const MIN_LENGTH = 1;
-const MAX_LENGTH = 125;
+const tokenizeWords = require('talisman/tokenizers/words/gersam');
+
+// Minimum of words that qualify as a sentence.
+const MIN_WORDS = 1;
+
+// Maximum of words allowed per sentence to keep recordings in a manageable duration.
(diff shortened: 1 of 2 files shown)


server/lib/validation/languages/ca.js
```diff
@@ -0,0 +1,38 @@
...
+
+// Maximum of words allowed per sentence to keep recordings in a manageable duration.
+const MAX_WORDS = 14;
+
+const INVALIDATIONS = [{
+  fn: (sentence) => {
+    const words = tokenizeWords('ca', sentence);
+    return words.length < MIN_WORDS || words.length > MAX_WORDS;
+  },
+  error: `Number of words must be between ${MIN_WORDS} and ${MAX_WORDS} (inclusive)`,
```
MichaelKohler, line 14: I think it would be easier if those errors would be translated to Catalan. Would make it possible to contribute without having to know English. What do you think? Other languages have done this already, most recently for example `yue`.

MichaelKohler (changes requested):
Thanks for submitting this. Generally looks good to me, just 2 things to comment on.

jmigual:
Sure I'll solve it right away

jmigual:
Issues fixed

MichaelKohler (approved):
Thanks! This will be part of the next Sentence Collector release. Once that happens, there will be a new comment on this PR.

Issue #609, opened 2022-03-08
Review Sentences can get stuck on loading


Sobsz:
My Sentence Collector session entered a "stuck" state. In this state, the "Review" section perpetually displays a spinner with the text "Loading sentences...", even when no language is selected. All other sections work fine. Debugging shows that no request is ever made to `review?locale=[language]` is ever made.

**The "stuck" state can be removed** by editing the local storage key `persist:sentences` to remove `sentencesLoading:"true"`, but this cannot be done without access to a trusted computer. None of the following mobile-only methods solved the issue:
- logging out and back in (whether via Firefox account or passwordless login)
- force-stopping Firefox
- disabling all addons and Enhanced Tracking Protection
- switching from mobile data to Wi-Fi

Switching browsers or using Incognito mode do allow the user to sidestep the issue. Removing browsing data would solve, too, but I could not find a way to do so for one site only on Firefox Android.

Running Firefox 97.2.0 on Android 11. My Internet connection was being spotty today, which I believe might have caused the problem.

MichaelKohler:
Thanks for this thorough bug report and all the debugging done. This makes it fairly straightforward to fix.

We are persisting the sentences state with a few exceptions here: https://github.com/common-voice/sentence-collector/blob/main/web/src/reducers/index.ts#L21

To make sure we are not persisting the loading states, we should add both `isUploadingSentences` and `sentencesLoading` to the exceptions linked to above. Then saving the state while a request is reloading wouldn't result in this scenario.

Great catch!

PR #622, opened 2022-05-25
fix: stop persisting loading state

paulotokimatu:
In order to simulate the issue reported by #609 I added some artificial delay in the `/review` endpoint of the server and while the web app was trying to fetch the sentences I just had to change pages or refresh your browser. This will make the app stuck in the loading state.

As mentioned by @MichaelKohler, the easy solution is to not persist the loading state `sentencesLoading`.

The key `isUploadingSentences` is also a loading state, but I was not able to simulate any issues like above because it goes to the catch and the state goes back to `false`. I also added it in the exception list just for consistency sake, but I can remove if you all think it is better since it is not causing real problems for now.

I didn't add any tests since it this case would be better tested by some type of integration test, but doesn't look like we are using them in the repo, right?

### DIFF
--- a/web/src/reducers/index.ts
+++ b/web/src/reducers/index.ts
@@ -18,7 +18,7 @@ const languagesPersistConfig = {
 const sentencesPersistConfig = {
   key: 'sentences',
   storage,
-  blacklist: ['sentences'],
+  blacklist: ['isUploadingSentences', 'sentences', 'sentencesLoading'],
 };
 
 export default function (history: History) {


MichaelKohler (approved):
Thanks for contributing this PR. This looks good to me!

> I didn't add any tests since it this case would be better tested by some type of integration test, but doesn't look like we are using them in the repo, right?

That's correct, currently we do not have any integration tests at all.

PR #630, opened 2022-07-30
add korean sentence validation & cleanup

sftblw:
This adds korean validation and cleanup.

- Korean script should only allow korean letters and some special characters `[.,?!]` (in my opinion).
  - There are so many things to be not allowd, so I added the rule as whitelist rule.
  - For responsive message I added some rules before it, which may be common mistakes.
- For sentence length calculation, I replaced it with character length.
  - In my opinion, Tokenization is somewhat heavyweight for this task, especially in CJK (Chinese, Japanese, Korean).
  - max length 50 is somewhat random, but the famous pengram "키스의 고유조건은 입술끼리 만나야 하고 특별한 기술은 필요치 않다" is 35 length so... It would be sufficient.
- About normalization
  - Korean input system has (mainly) two kind of characters (in unicode space), which are interchangeable but have different unicode code point.
  - In my opinion both should be allowed to be submitted, but normalized into one kind of codepoint when It comes to actual usage or sentence storing in DB.
  - so I added normalization to both of validation and cleanup.

### DIFF
--- a/server/lib/validation/languages/ko.js
+++ b/server/lib/validation/languages/ko.js
@@ -1,0 +1,55 @@
+// Minimum of characters that qualify as a sentence.
+const MIN_CHARACTERS = 1;
+
+// Maximum of characters allowed per sentence to keep recordings in a manageable duration.
+const MAX_CHARACTERS = 50;
+
+const INVALIDATIONS = [{
+  fn: (sentence) => {
+    // To properly tokenize Korean, We need some heavy tokenizers (ex: mecab-ko, nori, ...),
+    // For counting letters those tokenizers are not necessary.
+    return sentence.length < MIN_CHARACTERS || sentence.length > MAX_CHARACTERS;
+  },
+  error: `문장의 글자 수는 ${MIN_CHARACTERS}글자 이상, ${MAX_CHARACTERS}글자 이하여야 합니다.`,
+}, {
+  // One Korean letter is composed with two or three letters,
+  // in order of (consonant(1st) - vowel(2nd) - consonant(3rd, optional)).
+  // It shouldn't be allowed to use them separately, since that could cause various pronunciation issues.
+  // 
+  // This regex is for Unicode "Hangul Syllables" (U+AC00–U+D7A3), Which are composed form (see below).
+  regex: /[ㄱ-ㅎㅏ-ㅣ]/,
+  error: '문장에는 자음이나 모음만 따로 있는 글자가 있어서는 안 됩니다.',
+},
+{
+  // Korean letters (Hangul) have two type of Unicode code points.
+  //
+  // - Composed form (Unicode "Hangul Syllables" : U+AC00–U+D7A3)
+  //   - One Unicode codepoint contains three or two letters in rectangular shape.
+  //   - This is normally used codepoints.
+  // - Other forms
+  //   - Other Unicode codepoints deal korean letters as separated vowels and consonants.
+  //   - This takes doubled space in bytes.
+  //   - This only appears when a contributor is using keyboard layout called "Sebeolsik", which is akin to Dvorak.
+  //   - After NFC normalization ( 5a86a81 ),
+  //     Composible combination of two or three characters (1st - 2nd - 3rd (optional)) will become
+  //     Composed form ("Hangul Syllables"). Characters that cannot be combined may remain.
+  //
+  // This regex is for codepoints other than "Hangul Syllables" (U+AC00–U+D7A3).
+  regex: /[\u1100-\u11FF\uA960-\uA97F\u3130-\u318F]/u,
+  error: '문장에는 첫가끝 형태의 분해된 글자가 있어서는 안 됩니다. 완성형 글자를 입력해주세요.',
+}, {
+  // Since there are so may kinds of "should not be allowd" letters,
+  // It would be convenient to allow only certain type of characters.
+  // examples: CJK chinese letters, Japanese letters, Korean specific chinese letters (aka hanja),
+  //           not-used symbols (semicolon, colon - native korean sentences do not contain them),
+  //           better to be excluded symbols (quote, tilda, ...),
+  //           characters that can be normalized into normal characters with destructive NFKC normalization (ⓐ, ㈜, ...),
+  //           historical korean letters (aka 옛한글 - ㆆ, ㅿ, ㆁ, ...)
+  //           ...
+  regex: /[^가-힣.,?! ]/u,
+  error: '문장에는 한글과 마침표, 쉼표, 느낌표, 물음표, 공백만 들어있어야 합니다.',
+}];
+
+module.exports = {
+  INVALIDATIONS,
+};
--- a/server/lib/validation/index.js
+++ b/server/lib/validation/index.js
@@ -7,6 +7,7 @@ const eo = require('./languages/eo');
 const ig = require('./languages/ig');
 const it = require('./languages/it');
 const kab = require( './languages/kab');
+const ko = require( './languages/ko');
 const ne = require('./languages/ne');
 const or = require('./languages/or');
 const ru = require('./languages/ru');
@@ -25,6 +26,7 @@ const VALIDATORS = {
   ig,
   it,
   kab,
+  ko,
   ne,
   or,
   ru,


server/lib/cleanup/languages/ko.js
```diff
@@ -0,0 +1,16 @@
...
+
+function sortSentences(sentences) {
+  return sentences.sort();
+}
+
+function clean(sentences) {
+  return sentences.map((sentence) => {
+    // Normalization in Korean letters are important,
+    // since there are many kinds of Unicode Character which are same letter.
+    return sentence.normalize('NFC');
```
MichaelKohler, line 14: I can't really comment on the normalization per se, I think @ftyers can provide more information here. Is this something that indeed would be valuable to  do?

I have one comment here though: the cleanup script is run when exporting to Common Voice. This means that the sentences within the Sentence Collector would not be normalized. Might this lead to confusion? Also given that the validation would be done on the normalized string instead of what will actually be shown in the Sentence Collector GUI?

server/lib/validation/languages/ko.js
```diff
@@ -0,0 +1,54 @@
+// Minimum of characters that qualify as a sentence.
+const MIN_CHARACTERS = 1;
+
+// Maximum of characters allowed per sentence to keep recordings in a manageable duration.
+const MAX_CHARACTERS = 50;
```
MichaelKohler, line 5: How long does it approximately take to speak out 50 characters out loud?

server/lib/validation/languages/ko.js
```diff
@@ -0,0 +1,54 @@
...
+const INVALIDATIONS = [{
+  fn: (sentence) => {
+    // To properly tokenize Korean, We need some heavy tokenizers (ex: mecab-ko, nori, ...),
+    // For counting letters those tokenizers are not necessary.
+
+    // Before counting letters, normalization should be applied, since length may be tripled
+    const normalizedSenetence = sentence.normalize('NFC');
+    return normalizedSenetence.length < MIN_CHARACTERS || normalizedSenetence.length > MAX_CHARACTERS;
+  },
+  error: `Number of characters must be between ${MIN_CHARACTERS} and ${MAX_CHARACTERS} (inclusive)`,
```
MichaelKohler, line 16: For some other validators we started to show the errors in the respective languages. Might be worth to do this here as well :)

server/lib/validation/languages/ko.js
```diff
@@ -0,0 +1,54 @@
...
+}, {
+  regex: /[0-9]+/,
+  error: 'Sentence should not contain numbers',
+}, {
+  regex: /[<>+*#@%^[\]()/]/,
+  error: 'Sentence should not contain symbols',
+}, {
+  regex: /[a-zA-Z]/,
+  error: 'Sentence should not contain latin alphabet characters',
+}, {
```
MichaelKohler, line 26: I think the three rules here can be removed, as the last rule in this list would catch those anyway. What do you think?

server/lib/validation/languages/ko.js
```diff
@@ -0,0 +1,54 @@
...
+  // One Korean letter is composed with two or three letters,
+  // in order of (consonant(1st) - vowel(2nd) - consonant(3rd, optional)).
+  // It should be not allowed to use them separately, since That could cause various pronounciation.
```
MichaelKohler, line 29: minor:
```suggestion
  // It shouldn't be allowed to use them separately, since that could cause various pronunciation issues
```

server/lib/validation/languages/ko.js
```diff
@@ -0,0 +1,54 @@
...
+  // It would be convenient to allow only certain type of characters.
+  // examples: CJK chinese letters, Japanese letters, Korean specific chinese letters (aka hanja),
+  //           not-used symbols (semicolon, colon - native korean sentences do not contain them),
+  //           better to be excluded symbols (quote, tilda, ...),
+  //           characters that can be normalized into normal characters with NFKC normalization (ⓐ, ㈜, ...),
+  //           historical korean letters (aka 옛한글 - ㆆ, ㅿ, ㆁ, ...)
+  //           ...
+  /** @param {string} sentence */
+  fn: (sentence) => {
+    // With non-destructive normalization, We could allow users to input "Hangul Jamo" (not composed letters in unicode, aka 첫가끝).
```
MichaelKohler, line 44: minor:
```suggestion
    // With non-destructive normalization, we could allow users to input "Hangul Jamo" (not composed letters in unicode, aka 첫가끝).
```

server/lib/validation/languages/ko.js
```diff
@@ -0,0 +1,54 @@
...
+    // This is convenient for whom using specific type of keyboard layout (sebeolsik), since these layout inserts Hangul Jamo.
```
MichaelKohler, line 45: minor:

```suggestion
    // This is convenient for contributors using a specific type of keyboard layout (sebeolsik), since this layout inserts Hangul Jamo.
```

MichaelKohler (commented):
Thanks for contributing these. I have added a few comments.

sftblw:
Thank you for feedback.

## About normalization

Normalization was intended for contributors who use certain kind of keyboard layout, called "Sebeolsik".

Korean letters (Hangul) in unicode has two kind of codepoints.

- Composed form (Unicode "Hangul Syllables" : U+AC00–U+D7A3)
  - One Unicode codepoint contains three or two letters in rectangular shape.
  - This codepoints are normally used.
  - ex) `col` is `콜` which is `U+CF5C`, taking only one codepoint.
- Other forms
  - Other Unicode codepoints deal korean letters as separated vowels and consonants.
  - This takes doubled space in bytes.
  - ex) `col` is `콜` and it is actually three codepoints, `ᄏ` (1st k) `ᅩ` (2nd oh) `ᆯ` (3rd r).
  - This only appears when a contributor is using keyboard layout called "Sebeolsik", which is akin to Dvorak, or It appears when it comes to filename of macOS, since macOS normalize filenames with decomposition normalization (NFD).
- Two kinds of codepoint can be converted by each with normalization.
  - `"콜".normalize('NFD').split('') ` is `Array(3) [ "ᄏ", "ᅩ", "ᆯ" ]`.
    - and `"콜".split('')` is `Array [ "콜" ]`, proving its one and only codepoint.
  - `[ "ᄏ", "ᅩ", "ᆯ" ].join('').normalize('NFC').split('')` is `Array [ "콜" ]`

It comes clear if the sentence is normalized with composition normalization (NFC) before the sentence is submitted to the database, but I couldn't find existing way to normalize(morph) the sentence before submitted. Cleanup script is added to complement this in other way while supporting minor cases (And I think that is not clear way to do, too.)

So, It might be handled with one of two ways:

1. Simply prevent contributors to input decomposed code points (only allowing "Hangul Syllables"), Since this make things clear and those kind of codepoints are quite rarely used (in my opinion).
2. Add normalization before submitting, changing original sentence (Since NFC is non-destructive normalization, it is perfectly fine to do).

(and `/[^가-힣.,?! ]/u` only allows composed form, It's my mistake.)

## Character length

I checked one of "Public for Korean" dataset from [aihub.or.kr](https://aihub.or.kr/aihubdata/data/view.do?dataSetSn=130), - validation set, random 10 sentences (without actually listening them, believing its metadata to be correct)

> validation - broadcast set

data no.	| len	| second
----|-------|-------
14	| 121 | 15.87
29	| 38	| 3.84
43	| 37	| 4.48
61	| 37	| 3.46
458	| 97	| 11.01
466	| 61	| 7.3
475	| 73	| 10.37
738	| 44	| 5.5
826	| 45	| 5.12
993	| 52	| 5.63

( 605 characters / 72.58 second) = 8.33563 character / second

MichaelKohler:
Thanks for the explanation. I'm now checking if we could instead add NFC somewhere else (even before validation) to make this simpler. I will report back.

HarikalarKutusu:
> max length 50 is somewhat random, but the famous pengram "키스의 고유조건은 입술끼리 만나야 하고 특별한 기술은 필요치 않다" is 35 length so... It would be sufficient.
+
>  8.33563 character / second
Wouldn't it better to increase the max sentence length? Something like 70?
Common Voice recordings are limited to 10 sec recordings, mainly optimized for batch processing in 8GB GPU's.
If you limit your max-length further, you may be forced to exclude many sentences, which are hard to come by.
I'm asking without knowing the language of course...

MichaelKohler:
I have checked with others and per-language NFC normalization is something we want to support. Therefore I implemented that capability here: https://github.com/common-voice/sentence-collector/commit/5a86a81a6da7533e9571d2411777ef058a4419bb . I've already enabled it for Korean, however it's not on the live website yet. I will create a new release once we have the validation here as well.

This means:

* You do not have to normalize before counting the characters, the sentence passed into the function will already be normalized
* The regex rules will be applied on the normalized sentence
* We do not need the cleanup changes

Would be great if you could update your branch to be based on the latest `main` branch, and test out my addition and whether this works correctly for your validation as well.

Happy to provide any more context if needed and thanks for the suggestion!

> Wouldn't it better to increase the max sentence length? Something like 70?
Common Voice recordings are limited to 10 sec recordings, mainly optimized for batch processing in 8GB GPU's.
If you limit your max-length further, you may be forced to exclude many sentences, which are hard to come by.
I'm asking without knowing the language of course...

Not by too much though. 70 would mean roughly 8.5 seconds on average. If somebody reads/speaks 20% slower than average, then it would already hit the 10 seconds hard limit.

MichaelKohler (changes requested):
Marking this as "request changes", see my previous comment.

sftblw:
Sorry, This PR close was mistake; I'm not familiar with forked Github repository and pressed the "sync" with discard button. I'll reopen after some modification.

sftblw:
- I made a new commit (because of the mistake)
- Changed error message to Korean language (If I understood the recommendation correctly)
- Last validation *will* cover above two validation, but I left them to guide contributors.
  - `ㅋㅋㅋ` means *lol* (laught sound) and those characters are composed form but only with 1st letter of three (or two).
  - Non-composed codepoints will not be covered with `[ㄱ-ㅎㅏ-ㅣ]`, since those codepoints are combined form. so Second one is yet needed (to guide contributors)
  - If we don't need guidance, that two validation rule is not required.

MichaelKohler (approved):
Thanks!

PR #651, opened 2022-11-27
Add validator for Lao (lo)

wannaphong:
Propose code for Lao language validator.

It's basic validator code.

### DIFF
--- a/server/lib/validation/languages/lo.js
+++ b/server/lib/validation/languages/lo.js
@@ -1,5 +1,5 @@
 // Lao rules
-// use any rule from Thai rules https://github.com/wannaphong/sentence-collector/blob/main/server/lib/validation/languages/th.js
+// use any rule from Thai rules https://github.com/common-voice/sentence-collector/blob/main/server/lib/validation/languages/th.js
 const MIN_LENGTH = 2;
 const MAX_LENGTH = 100;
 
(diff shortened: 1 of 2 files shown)


server/lib/validation/languages/lo.js
```diff
@@ -0,0 +1,34 @@
+// Lao rules
+// use any rule from Thai rules https://github.com/common-voice/sentence-collector/blob/main/server/lib/validation/languages/th.js
+const MIN_LENGTH = 2;
+const MAX_LENGTH = 100;
+
+const INVALIDATIONS = [{
+  fn: (sentence) => {
+    return sentence.length < MIN_LENGTH || sentence.length > MAX_LENGTH;
+  },
+  error: `Number of characters must be between ${MIN_LENGTH} and ${MAX_LENGTH} (inclusive)`,
```
MichaelKohler, line 10: As with other languages, I think it would make sense to translate these error messages into Lao. What do you think?

wannaphong: > As with other languages, I think it would make sense to translate these error messages into Lao. What do you think?

I agree with you. I will translate these messages into Lao.

MichaelKohler (approved):
Thank you for the contribution!

PR #664, opened 2023-02-17
Create am.js

quaja-abu:


quaja-abu:
please approve it

server/lib/validation/languages/am.js
```diff
@@ -0,0 +1,32 @@
+const tokenizeWords = require('talisman/tokenizers/words');
+
+// Minimum of words that qualify as a sentence.
+const MIN_WORDS = 6;
+
+// Maximum of words allowed per sentence to keep recordings in a manageable duration.
+const MAX_WORDS = 26;
```
MichaelKohler, line 7: Thanks for contributing this validation file! This generally looks good to me, though I have two questions:

* How long does it take on average roughly to speak out aloud 6 words?
* How long for 26?

Thanks!

quaja-abu: @MichaelKohler Thank you for your fast reply, and you asked me a very important question. to answer your questions.

- On average to speak out aloud 6 words takes 5 seconds, and
- On average speaking out aloud 26 words takes 15 seconds.
Thank you.

MichaelKohler: Thanks. In that case I would probably suggest 3-4 words minimum if reasonable sentences of that size exist.

For the maximum we have a limit on the Common Voice platform of 10 seconds for recordings, so I would suggest something around 8 seconds maximum to make sure that people speaking a bit slower won't get many of their contributions cut off.

quaja-abu: Oh if that is the case more than 16 words will exceed the maximum 10-second recording I guess. So, how about we make it 3 minimum and 15 maximum words?

MichaelKohler: Yes, I think that sounds reasonable to do. Thanks!

quaja-abu: OK. let me change the file then.

server/lib/validation/languages/am.js
```diff
@@ -0,0 +1,35 @@
...
+}, {
+  regex: /[<>+*#@%'"^[\]()/]/,
+  error: 'አረፍተነገሩ ውስጥ የተለዩ ምልክቶች ሊኖሩ አይገባም እባክዎ ያስተካክሉ',
+}, {
+  // Any words consisting of uppercase letters or uppercase letters with a period
+  // inbetween are considered abbreviations or acronyms.
+  // This currently also matches fooBAR but we most probably don't want that either
+  // as users wouldn't know how to pronounce the uppercase letters.
+  regex: /[A-Z]{2,}|[A-Z]+\.*[A-Z]+/,
+  error: 'Sentence should not contain abbreviations',
```
MichaelKohler, line 30: Thanks for the changes and the translations! I had forgotten to mention that but great that you noticed :)

Would you like to translate this error as well?

quaja-abu: I hope it is better now

MichaelKohler (approved):

PR #669, opened 2023-04-14
fix: escape quotes

moz-dfeller:


MichaelKohler (approved):
Oh no, didn't spot that. Thanks for fixing this. I can run the release later today if you don't get to it before me.

